PR #61, opened 2017-10-14
Implement password change feature on CLI

### LINKED ISSUE #56: Changing password on cli
Requesting feature to be able to change password via cli.

cryptokat:
resolves #56

### DIFF
--- a/src/main/java/org/semux/CLI.java
+++ b/src/main/java/org/semux/CLI.java
@@ -6,12 +6,12 @@
  */
 package org.semux;
 
-import java.io.Console;
 import java.io.File;
 import java.util.Arrays;
 import java.util.List;
 
 import org.semux.core.Wallet;
+import org.semux.core.WalletLockedException;
 import org.semux.crypto.EdDSA;
 import org.semux.crypto.Hex;
 import org.semux.utils.SystemUtil;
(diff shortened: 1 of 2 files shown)


src/main/java/org/semux/CLI.java
```diff
@@ -113,6 +120,8 @@ private static void printUsage() {
         System.out.println("                  list    List all accounts and exit");
         System.out.println("  -c, --coinbase  index   Specify which account to be used as coinbase");
         System.out.println("  -p, --password  " + "pwd     Password of the wallet");
+        System.out.println("  -p, --password  " + "pwd     Password of the wallet");
+        System.out.println("  --changepassword        Change password of the wallet");
```
semuxproject: Password change should be an account operation, suggest "--a changepassword"

src/main/java/org/semux/CLI.java
```diff
@@ -193,13 +202,33 @@ private static void listAccounts(Wallet wallet) {
...
 
+    private static void changePassword(Wallet wallet) {
+        if (password == null) {
+            password = SystemUtil.readPassword();
+        }
+        if (!wallet.unlock(password)) {
+            System.exit(-1);
+        }
+
+        Console console = System.console();
```
semuxproject: System console is not always available. Since you want to customize the prompt message, may overload the SystemUtil.readPassword() function.

src/main/java/org/semux/CLI.java
```diff
@@ -193,13 +202,33 @@ private static void listAccounts(Wallet wallet) {
...
+        String newPassword = new String(console.readPassword("Please enter the new password: "));
+
+        wallet.changePassword(newPassword);
```
semuxproject: Change password will fail if the wallet is not unlocked

src/main/java/org/semux/CLI.java
cryptokat: @semuxproject Isn't this a wallet-level operation?
`wallet.changePassword`

PR #79, opened 2017-10-17
Dockerfile to build semux cli container

cryptozero:


semuxproject:
Hi @cryptozero, thanks for the PR, but we're not going to merge docker support. The reason is that JVM is already an VM and there're no need to run it under another VM, docker container. We would prefer to encourage people to build single powerful validator node, rather than running many docker instances.

cryptozero:
Hello, docker container is not a full virtualized VM, the advantages are many:
- software + dependencies in same package (ie: especific version of java vm).
- Avoiding problems with software installed on host.
- No configuration drift on updates (redeploy).
- Backup independency from hardware.
- Fast deployment...

On the other hand keeping development and deployment separated is a valid point.

Thanks.

semuxproject:
Nice points. Thanks

Issue #80, opened 2017-10-18
add Button for Number of Votes 


phash:
Delegates Screen, right hand under "Acc #0 ..." dropdown
Add a Button for adding the maximum amount of available votes to vote or unvote.

PR #96, opened 2017-10-19
Add buttons for setting the max possible amount of votes (Issue: #80)

phash:
Added a button for setting the max possible amount of votes 
for selected account #80 

Added a button for setting the max possible amount of unvotes 
for selected delegate


semuxproject:
It's functionally correct but needs re-design. The input box get squeezed by default.

Suggest putting a small icon on the input box, instead of adding a button, as this functionality is not very critical.

error

phash:
The TextField has a prefered and a minimal size... should not be smaller than this, if I understand correctly..
I cannot re enact the situation, where it squeezes

semuxproject:
I'm still not very satisfied with the UX; will re-visit this PR after issue #112 is fixed.

phash:
damn.... this did not merge correctly... :(

PR #97, opened 2017-10-19
Fix NumberFormatExceptions on parsing doubles from inputFields

phash:
Localized double values use different decimal seperators. 
Double.parse(str) can't handle this.
Fixed it with using NumberFormat.parse(str).doubleValue()

### DIFF
--- a/src/main/java/org/semux/gui/SwingUtil.java
+++ b/src/main/java/org/semux/gui/SwingUtil.java
@@ -15,6 +15,7 @@ import java.awt.image.BufferedImage;
 import java.net.URL;
 import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
+import java.text.NumberFormat;
 import java.util.Comparator;
 import java.util.EnumMap;
 import java.util.Map;
@@ -23,6 +24,7 @@ import javax.swing.*;
 import javax.swing.table.DefaultTableCellRenderer;
 import javax.swing.table.TableColumnModel;
 import javax.swing.text.DefaultEditorKit;
+import javax.swing.text.NumberFormatter;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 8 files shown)


src/main/java/org/semux/gui/SwingUtil.java
```diff
@@ -195,22 +197,42 @@ public static JTextField editableTextField() {
...
     public static final Comparator<String> NUMBER_COMPARATOR = (o1, o2) -> {
-        return Double.compare(Double.parseDouble(o1), Double.parseDouble(o2));
+        try {
+            return Double.compare(NumberFormat.getInstance().parse(o1).doubleValue(),
+                    NumberFormat.getInstance().parse(o2).doubleValue());
+        } catch (ParseException e) {
+            logger.error("NumberFormatException", e);
+        }
+        // we should never come here -> work or exception
+        return 0;
```
semuxproject: `important` this is a library class; if you don't know how to deal with exception, you should throw instead of catch and return. From the caller's perspective, they got 0 return which would mean the two parameters are equal, while there is actually an exception.   YOUR COMMENTS ARE WRONG!

Also, your log error is not descriptive; exception name and message will be printed out by SLF4J. Did you test the exception case?

Similar issues below.

src/main/java/org/semux/gui/panel/SendPanel.java
```diff
@@ -166,15 +172,27 @@ public void setTo(byte[] addr) {
     }
 
     public long getAmount() {
-        return (long) (Unit.SEM * Double.parseDouble(amount.getText().trim()));
+        try {
+            return (long) (Unit.SEM * NumberFormat.getInstance().parse(amount.getText().trim()).doubleValue());
+        } catch (ParseException e) {
+            logger.error("NumberFormatException", e);
+        }
+        // we should never come here -> work or exception
```
semuxproject: Again, exception handling is problematic.

phash: I changed this.

Imho getFee() and getAmount() should handle their Exceptions, no need to throw it to ActionListener
Show a errorMessage for the User instead

src/main/java/org/semux/gui/SwingUtil.java
phash: Throws a unchecked NumberFormatException 
Contract of functionalInterface does not permit to throw checkedExceptions.

phash:
the way you set the values for TextFields is wrong. You use String.format - but should use NumberFormat and DecimalFormat.
I changed the TextFields for inserting Values to send to JFormattedTextField which will no longer allow nonNumberInputs.

phash:
Changed delegates input to allow only integer values

PR #128, opened 2017-10-23
Add Selection for my or all delegates

### LINKED ISSUE #111: Regular user mode- add "My Delegates" view
As a regular user that votes for delegates I should be able to view relevant delegates only. 
For example - I voted for 3 delegates out of active 50, and I'd like to view only delegates that I voted for. 
On this view I'd like to see the following:

1. My address from where I voted. 
2. Amount of my votes. 
3. Delegate address where vote was applied. 
4. Delegate reward for voting. 
5. Delegate status (online/offline)

Feel free to rework this item to anything more useful.

phash:
Added RadioButtons to select either (resolves #111)
- only my Delegates I already voted for
- all delegates


semuxproject:
After choosing the `show my delegate`, only one delegate is displayed (which is correct). After a while, click anywhere will pop up a warning dialog `You didn't vote ....` (wrong). More tests needed.

P.S> `My delegate` sounds like the delegates you own, not the delegates you voted.

PR #134, opened 2017-10-24
Max votes buttons for delegates

phash:
Cleaner approach to the MaxVotesButtons


semuxproject:
plus

phash:
why does this look that strange in your UI? I can't reproduce this...
I changed it to icons...

PR #137, opened 2017-10-24
Add Traditional Chinese Translation

cryptokat:
Tested under Debian 8 GNOME

semux_tc


testzcrypto:
Thanks! Please join Semux translation project here: https://crowdin.com/project/semux
I already import your translation, and will give you proofreader permission after you join.

PR #147, opened 2017-10-26
Fill in max possible Votes automatically

phash:
Fields get filled on PanelStart, delegates-click or account selection
change

another approach to #134

### DIFF
--- a/src/main/java/org/semux/gui/panel/DelegatesPanel.java
+++ b/src/main/java/org/semux/gui/panel/DelegatesPanel.java
@@ -36,6 +36,7 @@ import org.semux.core.Delegate;
 import org.semux.core.PendingManager;
 import org.semux.core.Transaction;
 import org.semux.core.TransactionType;
+import org.semux.core.Unit;
 import org.semux.core.state.DelegateState;
 import org.semux.crypto.Hex;
 import org.semux.gui.Action;
@@ -100,6 +101,7 @@ public class DelegatesPanel extends JPanel implements ActionListener {
                 JTable table = (JTable) me.getSource();
                 Point p = me.getPoint();
                 int row = table.rowAtPoint(p);
+                refreshMaxUnvotes();
                 if (me.getClickCount() == 2) {
                     Delegate d = tableModel.getRow(table.convertRowIndexToModel(row));
                     if (d != null) {
@@ -320,11 +322,13 @@ public class DelegatesPanel extends JPanel implements ActionListener {
 
         switch (action) {
         case REFRESH: {
+            refreshMaxVotes();
             refreshAccounts();
             refreshDelegates();
             break;
         }
         case SELECT_ACCOUNT: {
+            refreshMaxVotes();
             refreshDelegates();
             break;
         }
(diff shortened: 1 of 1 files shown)


src/main/java/org/semux/gui/panel/DelegatesPanel.java
```diff
@@ -320,11 +322,13 @@ public synchronized void actionPerformed(ActionEvent e) {
...
         switch (action) {
         case REFRESH: {
+            refreshMaxVotes();
             refreshAccounts();
             refreshDelegates();
             break;
         }
         case SELECT_ACCOUNT: {
+            refreshMaxVotes();
             refreshDelegates();
```
semuxproject: Add ``refreshMaxUnvote()`` here because the max unvote amount depends on both delegate and account.

src/main/java/org/semux/gui/panel/DelegatesPanel.java
```diff
@@ -320,11 +322,13 @@ public synchronized void actionPerformed(ActionEvent e) {
 
         switch (action) {
         case REFRESH: {
+            refreshMaxVotes();
```
semuxproject: Needs to add `refreshMaxUnvote()` because the account list may have been changed on the `REFRESH` event.

src/main/java/org/semux/gui/panel/DelegatesPanel.java
```diff
@@ -491,9 +495,30 @@ private void refreshDelegates() {
...
     }
 
+    /*
+     * updates textVote with the highest possible amount of SEM for this account
+     */
+    private void refreshMaxVotes() {
+        if (null != getSelectedAccount()) {
+            Long voteable = ((getSelectedAccount().getBalance() - Config.MIN_TRANSACTION_FEE_SOFT) / Unit.SEM);
+            textVote.setText(voteable > 1 ? String.valueOf(voteable) : "");
+        }
```
semuxproject: How about clear the text field if no account is selected? Also, consider use `SwingUtil.formatVote(long)`.

src/main/java/org/semux/gui/panel/DelegatesPanel.java
```diff
@@ -491,9 +495,30 @@ private void refreshDelegates() {
...
+
+    /*
+     * updates textUnvote with the amount of votes already voted for this delegate
+     */
+    private void refreshMaxUnvotes() {
+        if (getSelectedDelegate() != null)
+            textUnvote.setText(getSelectedDelegate().getVotesFromMe() / Unit.SEM > 0
+                    ? String.valueOf(getSelectedDelegate().getVotesFromMe() / Unit.SEM)
+                    : "");
+    }
```
semuxproject: IMPORTANT: the `getVotesFromMe()` is the total votes from all my accounts; should use `getDelegateState().getVote(byte[] voter, byte[] delegate)`.

Also, same issue as `refreshMaxVotes()`

phash: changed it - didn't realize it is for all votes of my account

phash:
done

semuxproject:
Did you test your code. Vote/unvote fields are filled with zeros.

phash:
I tested it - and here it looks fine... If I click a delegate I get the votes into my unvote Field

phash:
image

phash:
found an error - SwingUtil.formatVote() works in a different way now, so I had to calculate with full SEM and no longer divided by Unit.SEM

PR #178, opened 2017-11-06
adding message for invalid amounts 

lodyman93:


semuxproject:
Hi @lodyman93, how is the message being used?

lodyman93:
i created another two pull requests on voting panel and on send panel that utilize it

PR #187, opened 2017-11-06
Check if balance is high enough to unvote

### LINKED ISSUE #186: Voting with max semux cause unable to devote
I gave some user 1534.95 vote (i had 1535 and 0.05 was the fee) now i can't devote. It shouldn't let people to max their votes or at least give a warning about devote.

phash:
resolves #186
just gives a warning to the user

### DIFF
--- a/src/main/java/org/semux/gui/panel/DelegatesPanel.java
+++ b/src/main/java/org/semux/gui/panel/DelegatesPanel.java
@@ -349,6 +349,16 @@ public class DelegatesPanel extends JPanel implements ActionListener {
             } else if (value == 0L) {
                 JOptionPane.showMessageDialog(this, MessagesUtil.get("VotesGreaterThanZero"));
             } else {
+                // Only a warning!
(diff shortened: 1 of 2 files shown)


src/main/java/org/semux/gui/panel/DelegatesPanel.java
```diff
@@ -349,6 +349,16 @@ public synchronized void actionPerformed(ActionEvent e) {
             } else if (value == 0L) {
                 JOptionPane.showMessageDialog(this, MessagesUtil.get("VotesGreaterThanZero"));
             } else {
+                // Only a warning!
+                if (value < Config.MIN_TRANSACTION_FEE && value >= 0L) {
```
semuxproject: I didn't get it. You're checking value instead of (balance - value - fee > Config.MIN_TRANSACTION_FEE) if it's VOTE

phash:
you are right - the calculation was incorrect

PR #205, opened 2017-11-10
Add AddressBookFeature

### LINKED ISSUE #138: Adressbook missing
Is there need of an adress book?

Stored locally maybe

phash:
resolves #138
added an AddressBook:
- AddressBookPanel added to MainFrame
- AddressBookChanges possible
- Add to AddressBook on SendPanel and Delegates Panel


semuxproject:
Hi @phash, thank you but I'm afraid that we're not going to merge this PR at the moment.

1. The address book feature is not a core function, and thus should not be in the toolbar. IMO this is only useful in the Send panel when we want to transfer coins to an known address. In bitcoin-qt, the address book was also moved from toolbar to an icon in the Send panel.
2. Got an exception when testing this functionality:
```
Exception in thread "AWT-EventQueue-0" java.lang.IndexOutOfBoundsException: Invalid index
	at javax.swing.DefaultRowSorter.convertRowIndexToModel(DefaultRowSorter.java:514)
	at javax.swing.JTable.convertRowIndexToModel(JTable.java:2642)
	at org.semux.addressbook.AddressBookPanel$1.mousePressed(AddressBookPanel.java:60)
```
3. This code is not concise and poorly organized; should be under `org.semux.gui.dialog` with one class rather than an independent package `org.semux.addressbook`.
4. There are no test cases covering the code. A lot of efforts are needed to maintain these 800+ lines of code after merging.
5. Add a new database into the kernel for a wallet function doesn't seem to be a good choice.

If you can re-implement this feature with careful design and simple, clean and well-tested code, we'll review it.

phash:
complete overhaul

the LoC Number comes from much boilerplate code of swing

phash:
Please take a look, I think it is better niw

Issue #206, opened 2017-11-11
Implement `--resync` option 


semuxproject:
Implement the resync feature

phash:
resync is: kill DB and go from 0?

semuxproject:
Not required before the database size hits 10+GB

PR #222, opened 2017-11-13
[PoC] Recover database in case of data corruption

cryptokat:
This is a PoC for https://github.com/semuxproject/semux/issues/206 based on the idea of removing all physical database files. All files under `database` folder will be removed from disk prior to Kernel initialization if `--resync` option is passed.
@semuxproject Please review whether the approach is appropriate.


semuxproject:
Updated the title and comments a bit.

semuxproject:
Close this PR as it seems to be a less prioritized feature.

PR #246, opened 2017-11-17
GUI: Refactor button construction

phash:
Updated and cleared the way Buttons are created

a nicer AddressBookButton


semuxproject:
This is a good idea!!!

Please sync with upstream, solve the conflicts and go through all the GUI code again.

You're also advised to squash your commits afterwards, https://github.com/semuxproject/semux/blob/master/.github/contributing.md

PR #253, opened 2017-11-19
GUI Add export private key feature

### LINKED ISSUE #238: Import/Export Private Key From UI
The feature of import/export private key is only available from CLI.
As a common user, the feature will be more accessible from UI as #226 requested.

phash:
resolves #238
cleaned up the mess


semuxproject:
Hi @phash, you can run `mvn formatter:format` to format your code.

Also, I got an message related exception when running the wallet, after checking out your PR.

phash:
I think this is it now ^^

PR #254, opened 2017-11-19
GUI import private key

phash:
Import private keys via GUI Menu


semuxproject:
Implemented in fc87efe5de5674875aff3d90bdb062b02197a56b

Issue #207, opened 2017-11-11
Logs are not redirected to file on Windows


semuxproject:

semuxproject:
Fixed

PR #256, opened 2017-11-20
Build fat jar & windows executable 

cryptokat:
The pull request makes following changes:

- build an executable [fat jar](https://imagej.net/Uber-JAR) using [maven-shade-plugin](https://maven.apache.org/plugins/maven-shade-plugin/)
- build windows executables using Launch4j
  - semux.exe : GUI wrapper
  - semux-cli.exe : CLI wrapper (for some reason the GUI wrapper doesn't work under cmd.exe, so an independent wrapper is necessary for windows users who want to use CLI features)
- modify release.sh script to create tar/zip ball for specific platform
  - windows
    - semux.exe
    - semux-cli.exe
  - unix
    - semux.jar
- replace slf4j-simple with [logback](https://logback.qos.ch/) for file logging (debug.log) which will be helpful for debugging reported issues


cryptokat:
From the perspective of QA, I still need to figure out how to properly automate the testing of Launch4j wrapper on appveyor. Currently I am testing manually using VirtualBox and [MSIE VM](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/).

cryptokat:
Screenshots:

screenshot from 2017-11-20 23-34-00
screenshot from 2017-11-20 23-34-11

cryptokat:
I believe this pull request will also resolve #207

semuxproject:
Hi @cryptokat, how about macOS? I saw you removed semux.sh. There is no way for me to start other than type java...

semuxproject:
Also are JVM parameters supported with the executable?

cryptokat:
Hi @semuxproject I will add OSX bundler later. For now the GUI application works well by double-clicking the executable semux.jar of unix distribution. As you know CLI doesn't work with the executable semux.jar unless it's loaded through java command which will also be addressed.

cryptokat:
Remaining Tasks

- [ ] OSX: integrate https://bitbucket.org/infinitekind/appbundler into build process
- [ ] OSX: ensure that the CLI application can be executed
- ensure that JVM parameters can be tuned
  - [ ] windows
  - [ ] unix
  - [ ] osx
- [ ] QA: test launch4j wrapper on appveyor 
- [ ] QA: test osx build on travis-ci

semuxproject:
Suggest adding `<maxHeapPercent>80%</maxHeapPercent>`

cryptokat:
@semuxproject JVM memory options can no longer be tuned if `<maxHeapPercent>80%</maxHeapPercent>` is added to launch4j build options. An alternative way is to add `-XX:MaxHeapFreeRatio=80` into the default l4j.ini for specifying the maximum free ratio. 
https://github.com/semuxproject/semux/pull/256/commits/e5c7c44427e0e022ccac63f85b321421f61c3f7c

semuxproject:
The maxheapfreeratio won't affect the max heap JVM can allocate. The xmx is a declared max heap size. 

The challenge is that jvm by default set the xmx to 1/4 of physical memory or 1GB, whichever is larger.

This causes a problem if we don't specify the value. Wallet crashes when high load even though they have a large enough memory.

Can the xmx be override in l4j?

cryptokat:
@semuxproject Sorry I got it wrong.
According to https://github.com/mirror/launch4j/blob/master/head_src/head.c#L705 , maximum heap size of 32-bit runtime will always be capped to 1024mb regardless `maxHeapPercent`. However users can still override `-Xmx` decided by Launch4j by adding `-Xmx=4096m` into `semux.l4j.ini` file.

Here are the launch4j debugging logs from binary built with `<maxHeapPercent>80</maxHeapPercent>`: 

**64-bit**
```
Runtime used:	1.8.0_151 (64-bit)
Heap -Xmx:	Requested 0 MB / 80%, Available: 2148 MB, Heap size: 1719 MB
Launcher args:	-Xmx1719m
```

**32-bit**
```
Runtime used:	1.8.0_151 (32-bit)
Heap limit:	Reduced 2053 MB heap size to 32-bit maximum 1024 MB
Heap -Xmx:	Requested 0 MB / 80%, Available: 2567 MB, Heap size: 1024 MB
Launcher args:	-Xmx1024m
```

semuxproject:
We need to look into this. Theoretically, 32-bit machine has a physical memorry limit of 3.5GB. It should be possible to set xmx larger than 1g.I don't know why l4j has a hard cap of 1G. I will take a look at the source code laterThe reason I prefer percentage is that it can be adopted to different system configuration.Also, check systemutil.bench() in develop branch :)---- On Tue, 21 Nov 2017 11:34:12 -0500  cryptokat<notifications@github.com> wrote ----In fact 32-bit users will run into a fatal exception by specifying a Xmx greater than 1024m. We may have to only provide 64-bit executable if it is necessary to allocate more than 1024m memory space.   —You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or mute the thread.

cryptokat:
Progress: built osx app bundle and dmg
screen shot 2017-11-21 at 23 20 51

cryptokat:
According to this documentation, the maximum heap size ranges from 1.4G to 1.6G on 32-bit Windows
http://www.oracle.com/technetwork/java/hotspotfaq-138619.html#gc_heap_32bit

semuxproject:
I think it's better to support both 32-bit and 64-bit for windows; 1G is fine for regular users.

However, only 64-bit machines are allowed to participate in consensus.

cryptokat:
@semuxproject agree with that we should deliver both 32-bit and 64-bit executable for windows users

semuxproject:
I'm going to merge PR with the following changes,  for an emergent release

* drop CLI support for windows
* add temporary macOS distribution (copy from unix)

We should continue working on this, especially the remaining tasks

cryptokat:
@semuxproject Thanks for the update. OSX packaging can be a little more complicated, so I think it will deserve an independent pull request. Currently I am considering using `brew` for CLI distribution and `brew cask` for app bundle distribution on OSX platform.

PR #268, opened 2017-11-25
Add Wrapper Class

cryptokat:
### Motivation

The latest version of Semux consensus requires validators to have at least 3.2GB of memory space assigned to JVM. Linux users may not be aware of the requirement or have difficulties with modifying the shell script that bootstraps Semux to increase the assigned memory space.

### Implementation

Wrapper.java is a process wrapper mainly created for dynamic allocation of the maximum heap size of Semux JVM. The wrapper dynamically allocates 80% of available physical memory to Semux JVM when -Xmx does not appear in `--jvm-options`.


semuxproject:
I'd suggest renaming wrapper to Semux.java as it's gonna be the main entrance

cryptokat:
@semuxproject I am not sure whether the renaming will be appropriate because I have no plan to use this wrapper on Windows platform in place of Launch4j.

cryptokat:
@semuxproject 
I just saw issue https://github.com/semuxproject/semux/issues/275
In consideration that Launch4j doesn't provide dynamic checking on system conditions, it will make more sense to use the wrapper across all platforms including Windows. I will see what I can do.

semuxproject:
I don't think we need to check the machine conditions in the launcher, although the launcher has to make sure  the JVM can utilize the max system capability.

The validator requirements is only enforced only when the current node is a validator nodes. The majority of users are non-validator eventually. :)

Now, I'm enhance the `SystemUtil.bench()` method, which will print out more info to help validators diagnose the problem.

cryptokat:
@semuxproject I got it. The pull request should be pretty much done if we are not going to use the 
 wrapper on windows platform.

src/test/java/org/semux/wrapper/WrapperTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+@RunWith(PowerMockRunner.class)
+@PowerMockRunnerDelegate(Parameterized.class)
+@PrepareForTest({ ProcessBuilder.class, Process.class, Wrapper.class, SystemUtil.class })
+public class WrapperTest {
+
+    @Parameterized.Parameters
+    public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][] {
+                { new String[] { "--gui", "--jvm-options", "-Xmx1G -Xms1G" },
```
semuxproject: How about JVM options which starts with `--`? Will be interpreted by the CLI parser?

src/main/resources/logback.xml
```diff
@@ -21,7 +24,9 @@
     <logger name="org.semux.db" level="DEBUG" />
```
semuxproject: I just realized that all the default log filter level has been changed to DEBUG from INFO by default.  We should revert this.

https://github.com/semuxproject/semux/blob/bac842094cda6de0454a3a2315e867b208ecfefe/config/simplelogger.properties

src/main/java/org/semux/wrapper/WrapperConsoleLogFilter.java
```diff
@@ -0,0 +1,29 @@
...
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.core.filter.Filter;
+import ch.qos.logback.core.spi.FilterReply;
+
+/**
+ * The log filter prevents outputs of wrapper classes from being printed to
+ * console
+ */
+public class WrapperConsoleLogFilter extends Filter<ILoggingEvent> {
```
semuxproject: Add a special filter to the console appender for wrapper only doesn't seem to be elegant to me. See also #278

src/test/java/org/semux/wrapper/WrapperTest.java
cryptokat: There is no JVM option starting with `--` according to https://docs.oracle.com/javase/8/docs/technotes/tools/unix/java.html

PR #272, opened 2017-11-26
CLI change messages to i18n

phash:
Using MessagesUtil.get(..) for messages on the CLI


semuxproject:
I will fix the Scrutinizer issues. 

@cryptokat  your opinion on CLI localization?

src/main/resources/org/semux/gui/messages.properties
```diff
@@ -158,3 +158,27 @@ AddAddress = Add an address
 
 BackupFileExists = "{0}" already exists. Do you want to replace it?
 
+#cli
```
cryptokat: I think it will be more appropriate to move these CLI messages into an independent file because they are not related to the GUI package: `src/main/resources/org/semux/cli/messages.properties`

cryptokat:
Hi @phash 
FYI: I am working on improvements based on your branch in #281 in order to resolve some reported issues from Scrutinizer

Issue #278, opened 2017-11-27
Sperate system.out/err and SlF4J logs


semuxproject:
It's a little bit messy how the system output streams and logs are being used/displayed.

Also, user needs to be able to customize the log filter levels

Issue #274, opened 2017-11-26
Debug Logging - enhancement/configuration


ocvaad:
In current implementation log written to the single file. 
Should consider the following:

1. Configurable debug log size on UI or via config file. 
   1.1. First approach is to have one file with fixed size. 
    1.2 . Second approach is rename to *.old/ *.old1 /*.old2 etc.  log file once it gets to the maximum size 
     and start new log. 

2. Should consider implementation of different 'levels' for the debug output:
like: 
2.1. Basic - only critical messages + very basic connections info 
2.2. Advanced - critical messages + advanced connections info
2.3. Full - full console output to the log file. 

3. Should consider to extract messages with critical importance to separated log file.

cryptokat:
We may like to make the log rotating feature optional because some Linux users will prefer using the system-wide `logrotate`

cryptokat:
1st and 2nd points are addressed in #279

cryptokat:
@ocvaad addressed in #283

Issue #273, opened 2017-11-26
Refactor the Config


semuxproject:
All blockchain configurations are currently stored in the `Config.java` as mutable static variables. This has been complained by code analyze tools.

semuxproject:
Implemented

PR #279, opened 2017-11-27
Logback Config Improvements

cryptokat:
### Log level customization

Users will be able to customize log level by adding a JVM property `-Dlog.level=DEBUG`.

We may like to add a configuration of `log.level` into `config/semux.properties` as a standard way to configure the log level in the future. (blocked by #273)

### Add error.log

It will be easier to address error messages by having an error.log

### Issues

#278 
#274


semuxproject:
I don't think we should split the log into two files:

1. There are only a few ERROR logs;
2. Often we can't diagnose the problem form the error messages without context.

In terms of customizing the logging level, it's better to move/copy the logback.xml to `config` folder (and put it under classpath so it can be loaded)

PR #281, opened 2017-11-27
Scrutinizer Fix of MAJOR Issues

cryptokat:
### Packages

- [x] api
- [x] cli
- [x] consensus
- [x] core
- [x] crypto
- [x] db
- [x] gui
- [x] net
- [x] util
- [x] vm


cryptokat:
Got some issues regarding constants of Bytes class that final primitive arrays are not immutable. There doesn't seem to be a perfect solution prior to Java 9.

### Links

https://stackoverflow.com/questions/3700971/immutable-array-in-java

cryptokat:
Hmmm It seems like Scrutinizer-CI recognizes methods with changed signature as "new" elements

src/main/java/org/semux/core/Genesis.java
```diff
@@ -93,7 +93,7 @@ public static synchronized Genesis getInstance() {
                 }
 
                 // configurations
-                Map<String, Object> config = json.getJSONObject("config").toMap();
+                Map<String, Object> config = json.getJSONObject(GENESIS_DIR).toMap();
```
semuxproject: `GENESIS_DIR` is a little bit confusing to me

src/main/java/org/semux/core/exception/BlockchainImplException.java
```diff
@@ -0,0 +1,29 @@
+/**
+ * Copyright (c) 2017 The Semux Developers
+ *
+ * Distributed under the MIT software license, see the accompanying file
+ * LICENSE or https://opensource.org/licenses/mit-license.php
+ */
+package org.semux.core.exception;
+
+public class BlockchainImplException extends RuntimeException {
```
semuxproject: Better rename to `BlockchainException`?

src/main/java/org/semux/util/ApiUtil.java
```diff
@@ -7,10 +7,7 @@
 package org.semux.util;
 
 import java.io.IOException;
-import java.net.HttpURLConnection;
-import java.net.InetSocketAddress;
-import java.net.URL;
-import java.net.URLEncoder;
+import java.net.*;
```
semuxproject: My IDE will auto expand this :)

src/main/java/org/semux/util/ApiUtil.java
```diff
@@ -32,10 +27,6 @@ public ApiUtil(InetSocketAddress server, String username, String password) {
         this.password = password;
     }
 
-    public ApiUtil(String username, String password) {
-        this(DEFAULT_SERVER, username, password);
-    }
-
```
semuxproject: This method is useful to me. Why scrutinizer complains about it?

cryptokat: https://scrutinizer-ci.com/g/semuxproject/semux/code-structure/develop/java-class/org.semux.util.ApiUtil
```
Hardcoding an IP address (127.0.0.1) into your code is a bad practice as well as a security problem. Consider reading the address from another source like a property file.
```

semuxproject:
Also, `crowdin.yml` needs to be updated

cryptokat:
I am not touching Config.java because of #273

semuxproject:
That's fine. I'm working on refactoring the configs

PR #286, opened 2017-11-29
Increase Synchronization Performance by Caching EdDSAPublicKey

cryptokat:
### Motivation

According to profiling results, EdDSAPublicKey constructor consumes ~37% of CPU time of `EdDSAPublicKey::verify` mainly by [creating precomputed tables](https://github.com/str4d/ed25519-java/blob/master/src/net/i2p/crypto/eddsa/spec/EdDSAPublicKeySpec.java#L39). Considering that Semux has an infrequently changed list of validators, caching public keys can reduce the synchronization time significantly.

### Benchmark

The benchmark was running against the mainnet at the height of ~15K

#### Before 

selection_419

#### After

~42% saved CPU time on sync thread

selection_418

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -359,6 +359,13 @@
             <version>3.4.4</version>
         </dependency>
 
+        <!-- Google Guava -->
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>23.5-jre</version>
+        </dependency>
+
         <!-- Testing Libraries -->
         <dependency>
             <groupId>junit</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -359,11 +359,11 @@
             <version>3.4.4</version>
         </dependency>
 
-        <!-- Google Guava -->
+        <!-- Caffeine Caching Library -->
         <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <version>23.5-jre</version>
+            <groupId>com.github.ben-manes.caffeine</groupId>
+            <artifactId>caffeine</artifactId>
+            <version>2.6.0</version>
         </dependency>
 
         <!-- Testing Libraries -->
--- a/src/main/java/org/semux/consensus/SemuxSync.java
+++ b/src/main/java/org/semux/consensus/SemuxSync.java
@@ -44,6 +44,7 @@ import org.semux.net.ChannelManager;
 import org.semux.net.msg.Message;
 import org.semux.net.msg.consensus.BlockMessage;
 import org.semux.net.msg.consensus.GetBlockMessage;
+import org.semux.util.TimeUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -157,9 +158,7 @@ public class SemuxSync implements Sync {
             try {
                 exec.awaitTermination(1, TimeUnit.MINUTES);
                 end = Instant.now();
-                long durationSeconds = Duration.between(begin, end).getSeconds();
-                logger.info(String.format("Syncing finished, took %02d:%02d:%02d", durationSeconds / 3600,
-                        (durationSeconds % 3600) / 60, (durationSeconds % 60)));
+                logger.info("Syncing finished, took {}", TimeUtil.formatDuration(Duration.between(begin, end)));
             } catch (InterruptedException e) {
                 Thread.currentThread().interrupt();
                 logger.error("Executors were not properly shut down");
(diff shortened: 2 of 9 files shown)


pom.xml
```diff
@@ -359,6 +359,13 @@
             <version>3.4.4</version>
         </dependency>
 
+        <!-- Google Guava -->
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>23.5-jre</version>
+        </dependency>
+
```
semuxproject: What's the main reasons for introducing Guava? It's a fairly big library and has a lot of dependencies. We add library only when it's too hard to code over JDK or there isn't enough time to implement/test it :)

src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -303,7 +312,7 @@ private void process() {
                 try {
                     Thread.sleep(1000);
                 } catch (InterruptedException e) {
-                    // do nothing
+                    Thread.currentThread().interrupt();
```
semuxproject: I remember syncing is being executed by the [cons] thread. Not sure if this is correct; I need to double check.

src/main/java/org/semux/crypto/cache/EdDSAPublicKeyCache.java
```diff
@@ -0,0 +1,51 @@
...
+     * EdDSAPublicKey::verify mainly by creating precomputed tables. Considering
+     * that Semux has an infrequently changed list of validators, caching public
+     * keys can reduce the synchronization time significantly.
+     * <p>
+     * The cache is a concurrent hash map of Hex(byte[] pubkey) -> EdDSAPublicKey
+     * <p>
+     * softValues() allows GC to cleanup cached values automatically.
+     */
+    private static final ConcurrentMap<String, EdDSAPublicKey> pubKeyCache = CacheBuilder.newBuilder().softValues()
+            .<String, EdDSAPublicKey>build().asMap();
```
semuxproject: What's the max size of the cache? If unlimited, this is a potential vulnerability.

Did you considering LRUMap<>?

cryptokat: I have evaluated following types of cache:

1. LRUCache with maximumSize: this type of cache limits the maximum number of elements which can be stored in the LRUCache, once the limit is reached, the least-recently-used element will be removed
2. LRUCache with maximumWeight and: this type of cache is a LRUCache that calculates the capacity based on a customized weighted function instead of the number of elements. the type of cache can be useful for memory-size based cache.
3. weakValues(): this type of cache doesn't remove elements proactively like LRUCache, instead it creates WeakReference for JVM GC to decide when an element should be removed. 
4. softValues(): this type of cache is similar to weakValues(), except for that it creates a SoftReference rather than a WeakReference for each element.

#### LRUCache maximumSize or maximumWeight

Ideally we want to allow users to customize the size of cache according to their own system environment, but this is not that realistic when it comes to Java because the size of Java Object is not static. Special [instrumentation code](https://docs.oracle.com/javase/8/docs/api/java/lang/instrument/Instrumentation.html#getObjectSize-java.lang.Object-) has to be added for estimating object size by inspecting JVM at runtime which may lower the performance, also the result is not precise over time.

How about a static number of cached public keys? It will be hard to decide how many public keys we actually want to keep inside of the memory to acheive the maximum performance because public key verification will be executed against various of scenarios.

#### weakValues() or softValues()

weakValues() and softValues() allow JVM GC to recycle unused cache on-demand according to reference count and available JVM memory space, which will allow users to fully-utilize their allocated heap space. The decision of choosing softValues() is made based on that SoftReference will only be recycled when there is more memory space required by JVM, which makes it more like a LRU cache, while the maximum size of cache is up to end users.

By relying on JVM GC to recycle cached elements, apparently it will increase the load of garbage collector and the risk of crashing it.

### Conclusion

At the end I decide to use softValues() at a balance of performance and flexibility. I have tested the implementation against extremely low Xmx (64M) and found no serious issue.

Guava is introduced mainly for its implementation of Concurrent SoftReference HashMap.

### References

- SoftReference: https://docs.oracle.com/javase/8/docs/api/java/lang/ref/SoftReference.html
- WeakReference: https://docs.oracle.com/javase/8/docs/api/java/lang/ref/WeakReference.html
- getObjectSize: https://docs.oracle.com/javase/8/docs/api/java/lang/instrument/Instrumentation.html#getObjectSize-java.lang.Object-
- Plugging memory leaks with soft references: https://www.ibm.com/developerworks/library/j-jtp01246/index.html

cryptokat:
Hi @semuxproject 
I replaced Guava with [Caffine](https://github.com/ben-manes/caffeine/wiki/Benchmarks) which is an improved/indepedent version of Guava's CacheBuilder. I hope this will meet your standard :smile:

PR #287, opened 2017-11-29
Minimize semux.jar

cryptokat:
Enable `minimzeJar` feature of maven-shade-plugin to remove unused classes from semux.jar.
The size of semux.jar reduces from 15MB to 9MB on develop branch.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -54,6 +54,7 @@
                     </execution>
                 </executions>
                 <configuration>
+                    <minimizeJar>true</minimizeJar>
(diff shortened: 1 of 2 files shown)


semuxproject:
I don't think it's a safe step to move at this point. As the code is not 100% covered by tests, we don't know what classes would be loaded at runtime.

It also increases the risks from the libraries, as we don't know their implementation details.

cryptokat:
@semuxproject agree
In fact I am also worried about the same thing.

Issue #288, opened 2017-12-01
Trivial: add licence header and -Xmx4g note to linux startup scripts


phash:
Semux version: `e.g., v1.0.0-rc.3`
OS & Version: Linux
Commit hash : 6018cf9

the Linux-StartUpScripts are lacking licence headers and a note for the -Xmx4g option

semuxproject:
Since memory availability is automatically detected, there is no need for user to manually set the Xmx parameters.

PR #289, opened 2017-12-01
Add a note and license header to startup scripts

phash:
the startup scripts are lacking license headers
the -Xmx4g Option is explained

solves #288


cryptokat:
Hi @phash 
The process wrapper automatically allocates 80% of available memory on your system to Semux JVM. Additional JVM options have to be added as `./semux-cli.sh --jvm-options="-Xmx4G"`
Please see https://github.com/semuxproject/semux/pull/268

semuxproject:
We should not advertise the `Xmx` options. Improper value may cause the JVM to crash during runtime; automatically value selecting is much safer.

PR #292, opened 2017-12-02
Add IP Filter

### LINKED ISSUE #282: Implement network blacklist and whitelist
We need probably implement the IP blacklist and whitelist mechanism for potential DDoS attack, especially for validators.

One option would be adding two files, `whitelist` and `blacklist`, in the config folder in plain text format, so people update the list. The kernel will maintain an IP database in memory for fast querying, with serializing/deserializing supported.

API, and CLI potentially, needs to be enhanced to allow dynamic adding/removing IP entries.

Also, we need consider making connection policy configurable.

cryptokat:
to resolve #282

The pull request introduces an IP filter that users can define a list of ACCEPT/REJECT rules to build a flexible whitelist/blacklist.

#### Todos

- [x] add test cases for SemuxIPFilter
- [x] add test cases for SemuxIPFilter.Loader
- [x] ensure that javax.json doesn't break api endpoints
- [x] resolve conflicts with #293 
- [x] implement add whitelist api
- [ ] replace javax.json with [an alternative](https://blog.takipi.com/the-ultimate-json-library-json-simple-vs-gson-vs-jackson-vs-json/)


cryptokat:
Some example rules

#### Whitelist Local Network

```
{
  "rules": [
    {"type": "ACCEPT", "address": "127.0.0.1/8"},
    {"type": "ACCEPT", "address": "192.168.0.0/16"},
    {"type": "REJECT", "address": "0.0.0.0/0"}
  ]
}
```

#### Blacklist IP Addresses

```
{
  "rules": [
    {"type": "REJECT", "address": "1.2.3.4"},
    {"type": "REJECT", "address": "5.6.7.8"}
  ]
}
```

pom.xml
```diff
@@ -306,14 +306,30 @@
...
+            <version>1.4</version>
+        </dependency>
 
         <!-- JSON -->
         <dependency>
-            <groupId>org.json</groupId>
-            <artifactId>json</artifactId>
-            <version>20170516</version>
+            <groupId>javax.json</groupId>
+            <artifactId>javax.json-api</artifactId>
```
cryptokat: javax.json is introduced for its java 8 stream-style interface which produces more elegant code
see: https://javaee.github.io/jsonp/

src/main/java/org/semux/api/ApiHandlerImpl.java
```diff
@@ -81,49 +82,60 @@ public String service(String uri, Map<String, String> params, HttpHeaders header
...
             case ADD_TO_BLACKLIST: {
-                // TODO: blacklist
+                try {
+                    String ip = params.get("ip");
+                    if (ip == null || ip.trim().length() == 0) {
+                        return failure("Invalid parameter: ip can't be empty");
+                    }
 
-                return failure("Not implemented yet!");
+                    channelMgr.getIpFilter().blacklistIp(ip.trim());
```
cryptokat: @semuxproject Should we also take care of active connections here? End users might expect that an IP of an active peer gets blacklisted will also be disconnected.

src/test/java/org/semux/db/MemoryDB.java
```diff
@@ -1,128 +0,0 @@
-/**
```
cryptokat: MemoryDB is replaced with TemporaryDBRule in order to cover more methods of ApiHandler

cryptokat:
[javax.json](https://github.com/javaee/jsonp) is under GPL license which means we won't be able to distribute it along with our fat jar, therefore we need to decide on an alternative https://blog.takipi.com/the-ultimate-json-library-json-simple-vs-gson-vs-jackson-vs-json/

cryptokat:
Here's a more complete list of alternatives from https://github.com/akullpp/awesome-java#json

src/main/java/org/semux/net/ChannelManager.java
```diff
@@ -28,16 +31,30 @@
     private Map<InetSocketAddress, Channel> channels = new HashMap<>();
     private Map<String, Channel> activeChannels = new HashMap<>();
 
+    private final SemuxIpFilter ipFilter;
+
+    public ChannelManager() {
+        ipFilter = (new SemuxIpFilter.Loader())
+                .load(Paths.get(Constants.CONFIG_DIR, "ipfilter.json"))
+                .orElse(null);
```
semuxproject: You forget `Config.dataDir()`

src/main/java/org/semux/api/ApiHandlerImpl.java
```diff
@@ -43,34 +38,19 @@
...
     private Kernel kernel;
 
-    private Wallet wallet;
-    private Blockchain chain;
-    private ChannelManager channelMgr;
-    private PendingManager pendingMgr;
-    private NodeManager nodeMgr;
-
-    private Config config;
-
```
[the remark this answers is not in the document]
cryptokat: The fields are removed for improving testability. They are no longer mockable once they are stored as private fields. Please see testGetPeers as an example  https://github.com/semuxproject/semux/pull/292/files#diff-e59afdeb023668734778f2016801287dR139

semuxproject:
A few personal suggestions:
* The code is poorly documented, add javadoc to the public/protected methods;
* Got a few warnings: `serial version id`, `unused import`, `unused variables`; 
* The code is not concise, control the number of packages/files for each feature;
* If you basically lowercase the classname and get the variable names, this won't improve the readability;

src/main/java/org/semux/api/ApiHandlerImpl.java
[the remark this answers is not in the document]
semuxproject: Thanks for the clarification

cryptokat:
@semuxproject Thanks for the review. Could you please elaborate which part of code is not concise to  you?

semuxproject:
Your code quality is good; concise is not the right word. I was a little bit concerned about the package and classs explosion.

Don't adjust your code too much based on scrutinizer or similar tools. I liked their rating system at first, but don't after I found that the scores are only based on complexity, # of lines.

This is kernel code; it's supported to be complex. We only need to make sure it has no issues.

cryptokat:
@semuxproject okay thanks for the explanation. I will try to address this in the rewrite with the alternative JSON library.

cryptokat:
Someone built a framework for benchmarking all the JSON libraries: https://github.com/fabienrenaud/java-json-benchmark

cryptokat:
I will open another pull request for replacing the JSON library. This one is a little bit messed up by "merge with develop branch" commits.

cryptokat:
merged manually by @semuxproject

PR #312, opened 2017-12-10
Batch Payment

### LINKED ISSUE #93: Add batch payments feature
It would be very useful to add a new API call `/transfer_many`, which should accept `to` Array of recipients. So there is 1 input and many outputs within 1 transaction.

It can be used by pools and exchanges to process pending withdrawals and many other services which accept SEM from customers.

cryptokat:
To resolve #93 

The pull request adds a new transaction type `TRANSFER_MANY` which allows a single transaction to have multiple recipients.

### Todos

- [x] implement TransactionType.TRANSFER_MANY
- [x] implement /transfer_many API
- [x] limit the maximum number of recipients per transaction
- [x] integration tests
- [x] code cleanup
- [x] support multiple recipients on GUI

### Backward Compatibility

This is going to be a backward-incompatible change considering that `TRASFER_MANY` transactions will be rejected by previous versions of client.


semuxproject:
Is there gonna be a limit on the number of recipients?
How would this feature affect transaction limit, block size, and network propagation?

cryptokat:
Thanks for the reminder.
I think a limitation on the maximum number of recipients is necessary, otherwise a malicious user may perform a DoS attack by creating a transaction with a large number of recipients.

cryptokat:
Another problem is if we should apply the original minimum transaction fee to a transaction with multiple recipients. Applying the original minimum transaction fee (0.05 SEM) may lower the cost of a DoS attack. From the perspective of security, it would be preferred to set the minimum transaction fee to `0.05 SEM * number of recipients` to prevent malicious users from utilizing TRANSFER_MANY for a DoS attack.

cryptokat:
The maximum number of recipients per transaction would be `5110` by applying `100KB` as a limit of maximum transaction size. The number can be checked dynamically here: https://github.com/cryptokat/semux/blob/43bff839b827716cb07ff04db6c1622373c4ffe8/src/test/java/org/semux/core/TransactionTest.java#L67

cryptokat:
Consider the worst case of a DoS attack utilizing TRANSFER_MANY:
1. an attacker creates 5000 transactions each with 5110 recipients
2. the block size will be around 100KB * 5000 ~ 500 MB 
3. the cost of creating such a block will be 0.05 SEM * 5110 * 5000 = 1277500 SEM

cryptokat:
A benchmark will be provided to see whether the limit of 100KB per transaction will be viable or not.

cryptokat:
100 KB doesn't seem to be quite viable as it takes more than 5 minutes adding a block with 5000 transactions each with 5110 recipients. A number between 200 and 500 would be more practical for the current implementation of Blockchain. Further optimization on `BlockchainImpl` would be required for a higher throughput.

src/test/java/org/semux/bench/BlockchainPerformance.java
```diff
@@ -0,0 +1,90 @@
...
+    private static final byte[] coinbase = Bytes.random(30);
+    private static final byte[] prevHash = Bytes.random(32);
+    private static final EdDSA key = new EdDSA();
+    private static final long value = 20;
+    private static final long fee = 1;
+    private static final long nonce = 12345;
+    private static final byte[] data = Bytes.of("test");
+    private static final long timestamp = System.currentTimeMillis() - 60 * 1000;
+
+    private static void testLargeBlock(DBFactory dbFactory) {
```
cryptokat: A benchmark of the worst case adding 5000 TRANSFER_MANY transactions each with 399 recipients.
It takes 29 seconds adding such a block to blockchain database on my local machine.

cryptokat:
Tested GUI on devnet
selection_433

semuxproject:
I'd suggest limit the number of recipients to 256; more than that makes little sense.

The block size can't exceed 2MB. The max number of txs per block should be kept at 5000 (planed to decrease to 4096 at launch for robustness).

Each multiple transfer takes half_of_num_ricipients from the 5000 quota, since storage space is almost halved compared to sending one by one.

src/test/java/org/semux/bench/BlockchainPerformance.java
semuxproject: Block persistence is only small portion of network propagation

cryptokat:
Added an integration test sending 1000 SEM from 1 account to 2 accounts using TRANSFER_MANY: https://github.com/cryptokat/semux/blob/batch-payment/src/test/java/org/semux/integration/TransferManyTransaction.java

pom.xml
```diff
@@ -317,6 +317,18 @@
...
+            <artifactId>commons-io</artifactId>
+            <version>2.6</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+            <version>4.5.4</version>
+            <scope>test</scope>
+        </dependency>
```
semuxproject: I didn't see the immediate benefits of bringing in the httpclient library, even though it's for test only. Better write a simple utility class, or use the ApiClient, for your purpose.

src/main/java/org/semux/api/ApiHandlerImpl.java
```diff
@@ -195,9 +197,9 @@ private ApiHandlerResponse getPeers() {
      * @param params
      * @return result
      */
-    private ApiHandlerResponse addNode(Map<String, String> params) {
+    private ApiHandlerResponse addNode(Map<String, Object> params) {
```
semuxproject: Recommend keeping `Map<String, String>`. This introduces too many Java type-casting, which is generally not good practise.

src/main/java/org/semux/api/HttpHandler.java
```diff
@@ -56,6 +58,7 @@
 
     private static final int MAX_BODY_SIZE = 512 * 1024; // 512KB
     private static final Charset CHARSET = CharsetUtil.UTF_8;
+    private static final Pattern ARRAY_PARAM_PATTERN = Pattern.compile("\\[\\]$");
```
semuxproject: I assume you've done some research on this. `name[]=` is recommended by php scripts, but breaks our API compatibility. How about `name=v1,v2,v3`, and process it at the command handler?

*Code below is to be reviewed*

cryptokat:
@semuxproject Thanks for the review. Mentioned issues are fixed in in the latest commit.
The added integration test doesn't seem to be working on CI. Under investigation.

cryptokat:
Working on GUI support of TRANSFER_MANY from SendPanel. Users will be able to send a TRANSFER_MANY transaction by providing comma separated recipient addresses in `To` field of SendPanel.

cryptokat:
The implementation is done. There are test cases covering:

**Unit Tests**
- API: /transfer_many
- Core/TransactionExecutor: execution of TRANSFER_MANY transactions
- GUI/SendPanel: transfer an amount of SEM to multiple recipients

**Integration Test**

[The integration test](https://github.com/cryptokat/semux/blob/batch-payment/src/test/java/org/semux/integration/TransferManyTransactionIntegrationTest.java) starts up 4 nodes on dev net locally:
  - kernelValidator: registered as a delegate in genesis.json
  - kernelPremine: with 1 million SEM from height 0
  - kernelReceiver1: empty wallet
  - kernelReceiver2: empty wallet

Once the nodes are booted, kernelPremine will send a 1000 SEM `TRANSFER_MANY` transaction to kernalReceiver1 and kernelReceiver2. Then the test case will make sure that the transaction is broadcasted and processed by all nodes.

I will come back to this if I have ideas for more extensive tests. Please share your thoughts with me on how to test the feature more properly.

src/main/java/org/semux/api/ApiHandlerImpl.java
```diff
@@ -633,17 +640,21 @@ private ApiHandlerResponse doTransaction(Command cmd, Map<String, String> params
...
                     .withFee(params.get("fee"))
-                    .withData(params.get("data"))
-                    .build();
+                    .withData(params.get("data"));
+
+            if (type == TransactionType.TRANSFER_MANY) {
+                transactionBuilder.withToMany(Arrays.asList(StringUtils.split(params.get("to"), ",")));
+            } else {
+                transactionBuilder.withTo(params.get("to"));
+            }
```
semuxproject: Better clean this code. The `to` processing logic should be within the transaction builder, rather than the TransactionBuilder user has to to take care of this.

src/main/java/org/semux/api/ApiHandlerImpl.java
```diff
@@ -633,17 +640,21 @@ private ApiHandlerResponse doTransaction(Command cmd, Map<String, String> params
...
 
+            Transaction tx = transactionBuilder.build();
             if (kernel.getPendingManager().addTransactionSync(tx)) {
-                return new DoTransactionResponse(true, Hex.encode0x(tx.getHash()));
+                return new DoTransactionResponse(true, null, Hex.encode0x(tx.getHash()));
```
semuxproject: I saw a lot of `null` being passed to Response class. Is the message optional? if so, maybe create another constructor without `message` parameter.

src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -328,8 +328,8 @@ private boolean validateApplyBlock(Block block) {
         }
 
         // [2] check transactions and results
-        if (transactions.size() > config.maxBlockSize() || !Block
-                .validateTransactions(header, block.getTransactions())) {
+        if (transactions.stream().mapToDouble(Transaction::weightedSize).sum() > (double) config.maxBlockSize() ||
+                !Block.validateTransactions(header, block.getTransactions())) {
```
semuxproject: I was a little bit concerned using `double` to store weights. This may make this comparison non-deterministic and break consensus.

src/main/java/org/semux/core/Block.java
```diff
@@ -322,6 +322,15 @@ public long getTimestamp() {
...
     }
 
+    /**
+     * Get block size in bytes
+     *
+     * @return block size in bytes
+     */
+    public int size() {
+        return toBytesHeader().length + toBytesTransactions().length + toBytesResults().length + toBytesVotes().length;
+    }
```
semuxproject: Who is using `size()`. The `toBytesVotes().length` may be variant between nodes.

src/main/java/org/semux/core/Transaction.java
```diff
@@ -9,33 +9,49 @@
...
-    private TransactionType type;
+    /**
+     * The maximum number of recipients of a TRANSFER_MANY transaction
+     */
+    public static final int MAX_RECIPIENTS = 200;
+
+    /**
+     * The maximum length transaction data in bytes
+     */
+    public static final int MAX_DATA_LENGTH = 128;
```
semuxproject: Recommend renaming to `TRANSFER_DATA_MAX_SIZE`. `data` is generic and has different limit based on the type.

src/main/java/org/semux/core/Transaction.java
```diff
@@ -109,19 +132,35 @@ public Transaction sign(EdDSA key) {
...
+            return false;
+        }
+
+        if (numberOfRecipients() > MAX_RECIPIENTS) {
+            logger.warn(
+                    "ignoring large transaction (recipients: {}, hash: {})",
+                    numberOfRecipients, Hex.encode(getHash()));
+            return false;
+        }
+
```
semuxproject: This is probably not the best place to add logs. If the problematic transaction is from API/GUI (us), it should have been catched outside; if it's from the network (bad peer), it pollutes our log.  Transaction is not supposed to have application logic.

src/main/java/org/semux/core/Transaction.java
```diff
@@ -152,12 +191,42 @@ public TransactionType getType() {
...
      */
-    public byte[] getTo() {
-        return to;
+    public int numberOfRecipients() {
+        int toLength = to.length;
+        if (toLength % EdDSA.ADDRESS_LEN != 0) {
+            throw new TransactionException(
+                    "then length of 'byte[] to' array is not a multiple of " + EdDSA.ADDRESS_LEN);
+        }
+        return toLength / EdDSA.ADDRESS_LEN;
```
semuxproject: Again, no need to check.

src/main/java/org/semux/core/TransactionExecutor.java
```diff
@@ -83,13 +88,27 @@ public TransactionExecutor(Config config) {
...
+                long deduction = value * recipients.length + fee;
+
+                if (deduction <= available) {
+                    as.adjustAvailable(from, -deduction);
+                    for (byte[] recipient : recipients) {
+                        as.adjustAvailable(recipient, value);
+                    }
+                    result.setSuccess(true);
+                }
+                break;
```
semuxproject: `MAJOR`: balance is not checked before the transfer. This may break the state.

Also check if there is overflow or not.

src/main/java/org/semux/core/exception/TransactionException.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public TransactionException(Throwable cause) {
+        super(cause);
+    }
+
+    public TransactionException(String message, Throwable cause, boolean enableSuppression,
+            boolean writableStackTrace) {
+        super(message, cause, enableSuppression, writableStackTrace);
+    }
+}
```
semuxproject: Instead of using exception, I'm thinking of adding error code to the transaction result.

PR #317, opened 2017-12-12
GUI refactor UI classes: Panels

phash:
Single Responsobility Principle:
Classes should not implement ActionListener
Classes should not extend Components
rather they should have them as fields: 
Composition over inheritance

- [x] DelegatesPanel

- [x] HomePanel

- [x] ReceivePanel

- [x] SendPanel

- [x] TransactionsPanel

- [x] MainFrame

- [x] MenuBar

- [ ] Tests

- [ ] clean up


phash:
first private

PR #330, opened 2017-12-15
Extract validations and Listener from UI Class

phash:
refactor SendPanel
own class for ActionListenerImplementation
Validations.class for all Validations
Tests


src/main/java/org/semux/api/validation/Validations.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * validates the length of a memo message
+     * 
+     * @param memo
+     * @return true if memo is in bounds
+     */
+    public static boolean validateMemoSize(String memo) {
+        return null == memo || Bytes.of(memo).length <= Constants.MEMO_SIZE;
+    }
```
semuxproject: `memo` is deprecated name. Please use `data` instead

src/main/java/org/semux/gui/panel/SendPanel.java
```diff
@@ -365,4 +302,76 @@ public String toString() {
...
+            switch (action) {
+            case REFRESH:
+                refresh();
+                break;
+            case SEND:
+                sendAction();
+                break;
+            case CLEAR:
+                clear();
+                break;
```
semuxproject: Recommend renaming to `send()` for consistency.

src/main/java/org/semux/config/Constants.java
```diff
@@ -88,6 +88,16 @@
...
 
+    /**
+     * the maximum size of memos
+     */
+    public static final int MEMO_SIZE = 128;
+
+    /**
+     * length of address
+     */
+    public static final int ADDRESS_LENGTH = 20;
```
semuxproject: Address is already defined in `EdDSA`

src/main/java/org/semux/api/validation/Validations.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Validates the heigth of the fee
+     * 
+     * @param fee
+     * @param config
+     * @return
+     */
+    public static boolean validateFeeHighEnough(long fee, Config config) {
+        return fee >= config.minTransactionFee();
```
semuxproject: The fee validation rule will be changed soon.

phash:
redoing the changes

PR #334, opened 2017-12-17
Gui refactor action listener

phash:
Refactot the UI  for cleaner approach


phash:
I sometimes dislike git...

PR #340, opened 2017-12-18
Enable multi-recipient transfer transaction

cryptokat:
The feature enables batch-payment with up to 100K recipients per block.

**Benchmark result of BlockchainPerformance.java**

- Single-Recipient Block Size = 2097128 bytes, 8961 txs (1 recipient)
- Multi-Recipient Block Size = 2097148 bytes, 104834 recipients (1 tx)


semuxproject:
Can we postpone this feature?

- The main release is gonna be very soon, we don't have enough time to test; stability is prioritized
- There are few use cases where one needs to send different addresses the same amount of value
- Re-design is required for block explorers (they have to properly handle multi-transfer)

cryptokat:
Understood. Let's dealy the feature.

PR #342, opened 2017-12-18
Sort Pending Transactions By Fee & Timestamp in PendingManager

cryptokat:
Pending transactions should be sorted by:
- fee in in descending order
- timestamp in in ascending order


semuxproject:
Research and poc required

Can't simplely sort the executed transactions. Changing the order may affect the validity of the transactions

In butcoin, they have a main map and different views, like fee view and time view ,of pending transactions. Snapshots are created periodically when mining.

cryptokat:
Thanks for the review. I am working on further investigation of this feature.

PR #345, opened 2017-12-20
Report Cause of Rejection From PendingManager

cryptokat:
To make it less confusing for users when a transaction is rejected

- [x] report cause of rejection from `PendingManager`
- [x] display cause of rejection in api repsonse
- [x] display cause of rejection in GUI error dialog

### DIFF
--- a/src/main/java/org/semux/api/ApiHandlerImpl.java
+++ b/src/main/java/org/semux/api/ApiHandlerImpl.java
@@ -38,6 +38,7 @@ import org.semux.api.response.SendTransactionResponse;
 import org.semux.api.util.TransactionBuilder;
 import org.semux.core.Block;
 import org.semux.core.BlockchainImpl;
+import org.semux.core.PendingManager;
 import org.semux.core.Transaction;
 import org.semux.core.TransactionType;
 import org.semux.core.exception.WalletLockedException;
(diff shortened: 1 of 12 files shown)


semuxproject:
Awesome PR. 

There is only one potential issue with ProcessTransactionResult. Considering the following situation:

In the pending manager, we have one transaction in the cache (nonce jump)
tx1 = {nonce = 2, timestamp = now - two_hours + one_minute}

After two minutes, we receive another transaction
tx2 = {nonce = 1, timestamp = now}

We will get INVALID_TIMESTAMP because the `processTransaction()` will process tx2 (valid) and then tx1 (invalid,  expired) in the loop.

cryptokat:
Thanks for explaining the issue :smile: 
The issue is addressed in the latest commit.

semuxproject:
I don't think that's enough. The real issue is that we are not getting the error for the **current** transaction.

cryptokat:
Hi @semuxproject 
I don't really get your idea.
Checking tx timestamp at the beginning of `processTransaction` can prevent transactions with a large time drift from going into the `transactions`/`delayed` pool, so there is no reason to check tx timestamp once again for delayed transactions inside of the while loop.

semuxproject:
You're right. The timestamp check makes absolute sense.

The only thing that confused me is the error in the ProcessTransactionResult. I thought it's the error of the input transaction. After reading you documents, you said it's the `error that interrupted the process`. I think this also works.

I'm going to merge it. Thanks

cryptokat:
Cool. Thanks for the review!

Issue #353, opened 2017-12-22
OutOfMemoryError


jezzmemo:
Hi there,

Please note that this is an issue tracker reserved for bug reports and feature requests.

For general questions please use the Discord channel here: https://discord.gg/qQVckKZ

#### System information

Semux version: `1.0.0-rc.4`

OS & Version: `semux-macos`

#### Expected behaviour

None

#### Steps to reproduce the behaviour

1. Step one
Open application
2. Step two
Launch a few minutes

#### Backtrace

````
Exception in thread "gui-data" Exception in thread "msg-1" java.lang.OutOfMemoryError: GC overhead limit exceeded
java.lang.OutOfMemoryError: GC overhead limit exceeded
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.addWaiter(AbstractQueuedSynchronizer.java:606)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:1199)
	at java.util.concurrent.locks.ReentrantLock$NonfairSync.lock(ReentrantLock.java:209)
	at java.util.concurrent.locks.ReentrantLock.lock(ReentrantLock.java:285)
	at java.util.concurrent.ScheduledThreadPoolExecutor$DelayedWorkQueue.offer(ScheduledThreadPoolExecutor.java:1010)
	at java.util.concurrent.ScheduledThreadPoolExecutor$DelayedWorkQueue.add(ScheduledThreadPoolExecutor.java:1037)
	at java.util.concurrent.ScheduledThreadPoolExecutor$DelayedWorkQueue.add(ScheduledThreadPoolExecutor.java:809)
	at java.util.concurrent.ScheduledThreadPoolExecutor.reExecutePeriodic(ScheduledThreadPoolExecutor.java:346)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:296)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Exception in thread "pending-0" java.lang.OutOfMemoryError: GC overhead limit exceeded
````

semuxproject:
What's your system configuration, especially memory size?

jezzmemo:
[...]

PR #355, opened 2017-12-22
Wrapper: Set 128M as the minimum size of allocated memory to JVM

cryptokat:
In case of a busy system, an user may have an extremely low dynamically allocated memory to JVM of semux, which results in issues like #353. Setting the minimum size of allocated memory to JVM can enforce this kind of user to free up memory space for Semux.


semuxproject:
Recommend changing it to 1024

128 is too small :)

cryptokat:
changed

PR #361, opened 2017-12-27
Tools: add BitcoinjEnforcerRules to avoid dependency-chain attack

cryptokat:
see https://github.com/gary-rowe/BitcoinjEnforcerRules


semuxproject:
Nice to have this dependencies verification.

PR #363, opened 2017-12-27
Reject Transactions With Duplicated Hash

cryptokat:
Currently it's possible that an existing transaction gets overwritten in local blockchain database. The issue can be addressed by rejecting all existing transactions by looking up the `indexDb`.

#### References

- https://github.com/bitcoin/bips/blob/master/bip-0030.mediawiki
- https://github.com/bitcoin/bitcoin/commit/a206b0ea12eb4606b93323268fc81a4f1f952531

### DIFF
--- a/src/test/java/org/semux/consensus/SemuxBFTTest.java
+++ b/src/test/java/org/semux/consensus/SemuxBFTTest.java
@@ -78,24 +78,41 @@ public class SemuxBFTTest {
     @Test
     public void testDuplicatedTransaction() {
         // mock blockchain with a single transaction
-        EdDSA from = new EdDSA();
-        Transaction tx = new Transaction(
+        EdDSA to = new EdDSA();
+        EdDSA from1 = new EdDSA();
+        long time = System.currentTimeMillis();
+        Transaction tx1 = new Transaction(
                 TransactionType.TRANSFER, //
-                new EdDSA().toAddress(), //
+                to.toAddress(), //
                 10 * Unit.SEM, //
                 kernelRule.getKernel().getConfig().minTransactionFee(), //
                 0, //
-                System.currentTimeMillis(), //
+                time, //
                 Bytes.EMPTY_BYTES //
-        ).sign(from);
+        ).sign(from1);
         kernelRule.getKernel().setBlockchain(new BlockchainImpl(kernelRule.getKernel().getConfig(), temporaryDBRule));
-        kernelRule.getKernel().getBlockchain().getAccountState().adjustAvailable(from.toAddress(), 1000 * Unit.SEM);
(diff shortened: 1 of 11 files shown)


src/test/java/org/semux/consensus/SemuxSyncTest.java
```diff
@@ -0,0 +1,63 @@
...
+                Bytes.EMPTY_BYTES //
+        ).sign(from);
+        kernelRule.getKernel().setBlockchain(new BlockchainImpl(kernelRule.getKernel().getConfig(), temporaryDBRule));
+        kernelRule.getKernel().getBlockchain().getAccountState().adjustAvailable(from.toAddress(), 1000 * Unit.SEM);
+        Block block1 = kernelRule.createBlock(Arrays.asList(tx));
+        kernelRule.getKernel().getBlockchain().addBlock(block1);
+        SemuxSync semuxSync = new SemuxSync(kernelRule.getKernel());
+
+        // create a duplicated tx in the second block
+        Block block2 = kernelRule.createBlock(Arrays.asList(tx));
```
cryptokat: Actually there is no way to have such transaction in the network because of the implemented nonce system

src/test/java/org/semux/consensus/SemuxBFTTest.java
```diff
@@ -59,4 +71,50 @@ public void testIsPrimary() {
...
+        ).sign(from1);
+        kernelRule.getKernel().setBlockchain(new BlockchainImpl(kernelRule.getKernel().getConfig(), temporaryDBRule));
+        kernelRule.getKernel().getBlockchain().getAccountState().adjustAvailable(from1.toAddress(), 1000 * Unit.SEM);
+        Block block1 = kernelRule.createBlock(Arrays.asList(tx1));
+        kernelRule.getKernel().getBlockchain().addBlock(block1);
+        SemuxBFT semuxBFT = new SemuxBFT(kernelRule.getKernel());
+
+        // create a tx with the same hash with tx1 from a different signer in the second
+        // block
+        EdDSA from2 = new EdDSA();
```
cryptokat: Added a realizable scenario by signing the second tx with a different signer.

Since the signer of a tx doesn't affect the tx's hash, a malicious user may create duplicated tx hash using 2+ private keys.

src/test/java/org/semux/consensus/SemuxSyncTest.java
semuxproject: Awesome finding!

Let's also think about other scenarios where people can exploit this vulnerability.

semuxproject:
Side notes:
* block hash are secured by timestamp (monotonically increasing)
* coinbase transaction hash are secured by nonce (linearly increasing)

PR #367, opened 2017-12-27
API: avoid auth timing attack

cryptokat:


semuxproject:
> In cryptography, a timing attack is a side channel attack in which the attacker attempts to compromise a cryptosystem by analyzing the time taken to execute cryptographic algorithms.

Not sure whether this PR intends to solve the timing attack or to increase the password complexity.

Looks like timing attack is even easier since the server now has to do the hash work.

P.S.  Our API basic auth is also weak in the following aspects:
* The network traffic can't be captured;
* The configure file can't be stolen.

cryptokat:
This is inspired by CVE-2013-4165. The idea is to have a time-constant comparison between the user-provided password and the configured password in order to prevent attackers from guessing password of RPC by latency.

`MessageDigest.isEqual` is used in the latest commit: http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/tip/src/share/classes/java/security/MessageDigest.java#l430

### References

- bitcoin/bitcoin#2838
- bitcoin/bitcoin#2886
- https://codahale.com/a-lesson-in-timing-attacks/

PR #368, opened 2017-12-28
Limit number of nodes in NodesMessage 

cryptokat:
`NodeManager` may be vulnerable to OOM DoS when a malicious peer sends a large number of nodes through `NodesMessage`

### DIFF
--- a/src/main/java/org/semux/net/NodeManager.java
+++ b/src/main/java/org/semux/net/NodeManager.java
@@ -11,9 +11,8 @@ import java.net.InetSocketAddress;
 import java.net.UnknownHostException;
 import java.util.Collection;
 import java.util.HashSet;
-import java.util.Queue;
 import java.util.Set;
-import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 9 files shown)


src/main/java/org/semux/net/msg/p2p/NodesMessage.java
```diff
@@ -50,15 +57,15 @@ public NodesMessage(byte[] encoded) {
 
         nodes = new HashSet<>();
         SimpleDecoder dec = new SimpleDecoder(encoded);
-        int n = dec.readInt();
+        int n = Math.min(dec.readInt(), MAX_NODES);
```
semuxproject: How about enforcing the MAX_NODES rule in the handler, so that we can apply more logic after receiving this big nodesmessage?

Also, 128 seems too small to me since the validator nodes will increase to 100

src/main/java/org/semux/net/NodeManager.java
```diff
@@ -207,4 +216,38 @@ protected void doConnect() {
...
+        public Node(String hostname, int port) {
+            super(hostname, port);
+        }
+
+        public Node(InetAddress addr, int port) {
+            super(addr, port);
+        }
+
+        @Override
+        public int compareTo(Node o) {
```
semuxproject: I understand this is not necessary. To make the scrutinizer happy,  can we add an overriding equals() method which calls super.equals()?

PR #394, opened 2018-01-02
Avoid OOM caused by malicious BFT_NEW_HEIGHT messages

cryptokat:
`SemuxSync` may be vulnerable to OOM DoS when `SemuxBFT` receives a `BFT_NEW_HEIGHT` message with a large `height` property.

### DIFF
--- a/src/main/java/org/semux/consensus/SemuxBFT.java
+++ b/src/main/java/org/semux/consensus/SemuxBFT.java
@@ -428,11 +428,12 @@ public class SemuxBFT implements Consensus {
     protected void onNewHeight(long newHeight) {
         logger.trace("On new_height: {}", newHeight);
 
+        updateValidators();
         if (newHeight > height && activeValidators != null && !activeValidators.isEmpty()) {
             OptionalLong targetOptional = activeValidators.stream()
                     .mapToLong(c -> c.getRemotePeer().getLatestBlockNumber())
                     .sorted()
-                    .limit((int) Math.floor(activeValidators.size() * 2.0 / 3.0))
+                    .limit(Math.max((int) Math.floor(activeValidators.size() * 2.0 / 3.0), 1))
                     .max();
 
             if (targetOptional.isPresent() && targetOptional.getAsLong() > height) {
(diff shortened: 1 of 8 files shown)


semuxproject:
I don't think the `MAX_NEW_HEIGHT_DEVIATION ` will work for a brand new node to connect to the network.

IMHO, the logic for the newHeight should be:
```
if newHeight > height
    sort all known heights of validator nodes
    take the 2/3th height as target
    if target > height
        sync(target)
    endif
endif
```

cryptokat:
Good idea. I am working on the provided algo.

cryptokat:
Proposed algorithm is implemented in the latest commit.

src/main/java/org/semux/consensus/SemuxBFT.java
```diff
@@ -415,21 +416,28 @@ protected void jumpToView(int view, Proof proof, Proposal proposal) {
...
+     * 
+     * @param newHeight
+     *            new height
+     */
+    protected void onNewHeight(long newHeight) {
+        logger.trace("On new_height: {}", newHeight);
 
-            if (count >= (int) Math.ceil(activeValidators.size() * 2.0 / 3.0)) {
-                sync(h);
+        updateValidators();
```
semuxproject: I would not recommend refreshing validators here, since this method is invoked very often.

The active validator set is updated each time the node enters the PROPOSE state.

src/main/java/org/semux/consensus/SemuxBFT.java
```diff
@@ -415,21 +416,28 @@ protected void jumpToView(int view, Proof proof, Proposal proposal) {
...
+        if (newHeight > height && activeValidators != null && !activeValidators.isEmpty()) {
+            OptionalLong targetOptional = activeValidators.stream()
+                    .mapToLong(c -> c.getRemotePeer().getLatestBlockNumber())
+                    .sorted()
+                    .limit(Math.max((int) Math.floor(activeValidators.size() * 2.0 / 3.0), 1))
```
semuxproject: Better create a test case for different number of validators.

cryptokat:
Syncing with testnet after rebasing on top of develop branch:

selection_076

PR #423, opened 2018-01-05
Limit Inbound Connections of Each Unique IP Address

cryptokat:
Currently a single user from a single IP address can exhaust the quota `netMaxInboundConnections` of a node easily which implies a DoS attack. This pull request makes it harder for a malicious user to perform a network DoS attack by making idle connections.


src/main/java/org/semux/net/ConnectionLimitHandler.java
```diff
@@ -0,0 +1,78 @@
...
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+
+@ChannelHandler.Sharable
+public class ConnectionLimitHandler extends ChannelInboundHandlerAdapter {
+
+    private static final Logger logger = LoggerFactory.getLogger(ConnectionLimitHandler.class);
+
+    private static final ConcurrentHashMap<InetAddress, AtomicLong> connectionCount = new ConcurrentHashMap<>();
```
semuxproject: `hashCode()` of `InetAddress` is not implemented. Consider using `String` or alternatives

cryptokat:
The feature doesn't really work as expected on runtime. I will reopen the pull request once more verifications are done.

PR #424, opened 2018-01-05
GUI: OSX puts DB/wallet in ~/Library/Application Support fixes #260

### LINKED ISSUE #260: OSX Distribution
- [ ] Create .app bundle & .dmg image
  - https://bitbucket.org/infinitekind/appbundler
- [ ] Create homebrew formula
- [ ] QA: test osx build on travis-ci

orogvany:
Previously, the wallet and db were placed in the app itself.
Put it in the OSX expected location.  Log files are also
moved out of the app itself.  A template config is packaged
with the app and created if it does not exist.

OSX users like to have the .app as one atomic unit, so
external configuration should not be required or created
for them.

fixes #260

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -140,15 +140,16 @@
                 <artifactId>appbundle-maven-plugin</artifactId>
                 <version>1.2.0</version>
                 <configuration>
-                    <mainClass>org.semux.gui.SemuxGUI</mainClass>
+                    <mainClass>org.semux.gui.ApplicationLauncher</mainClass>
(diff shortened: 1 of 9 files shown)


semuxproject:
Decided to keep the original design for the first release; will consider the `user.home` or `Application Support` option later.

A simple fix is provided by 86a43d9f8fc3223a3963f0a963814b380b68e7ff

orogvany:
No worries!  It's a lot more stupid changes than I was hoping for OSX support, but worth exploring anyway :D  OSX is just a different beast with java to make it work like other apps.

Cheers

PR #434, opened 2018-01-06
Consensus: reject a block in the future

cryptokat:
A malicious validator may propose a block with a timestamp in the future which suspend the network indefinitely

### DIFF
--- a/src/main/java/org/semux/config/AbstractConfig.java
+++ b/src/main/java/org/semux/config/AbstractConfig.java
@@ -14,6 +14,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 
 import org.semux.core.TransactionType;
 import org.semux.core.Unit;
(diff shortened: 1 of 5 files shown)


orogvany:
Wouldn't even a jump of 2 hours still break the network?

The next validator to propose a block will have a date in the past, and will get denied, until the time matches up with that.  So the validator can control the block time for block following what they validate?

cryptokat:
The threshold is 15 minutes in this pull request which is adapted by Ethereum.

#### References

- https://github.com/ethereum/wiki/wiki/White-Paper#blockchain-and-mining

cryptokat:
A block proposal has to be validated by 2/3 of validators before it can be committed into the blockchain. This check adds a limitation on the maximum allowed time drift of a validator's system clock. Considering network propagation and desynced system clock, 15 minutes should be a reasonable threshold.

orogvany:
Oh, sorry, got the 2 constants's mixed up.
Think there's still a hole here, but I'll validate my hypothesis separately, and keep it out of the PR (once testnet is back up).
Thanks for clarifying! Sorry for confusion.

PR #449, opened 2018-01-08
API: Extract methods from handler into reusable location

orogvany:
Pull all the methods into an interface/callable methods.
This allows us to better support a command console like
other cryptos.  This could also be used later with the CLI.

SemuxAPI wasn't really an API, so it was renamed.

Also fix typo in Transaction builder.

### DIFF
--- a/config/semux.properties
+++ b/config/semux.properties
@@ -58,3 +58,10 @@ api.listenPort = 5171
 # Basic authentication
 api.username = YOUR_API_USERNAME
 api.password = YOUR_API_PASSWORD
+
+#================
+# UI
+#================
+
+# Specify the localization of UI
+# ui.locale = en_US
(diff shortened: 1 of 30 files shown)


semuxproject:
This would be a good enhancement.

@cryptokat  could you also help review this?

orogvany:
Thanks.  I have a console up and working for a second part to this.  

So you can open a window and type getInfo or getBlock 1234 etc.

Could add other methods in the future, and it'll auto update, but it's a starting point.

orogvany:
How can I get these going through?  I can pull just API into one PR, and then movign away in the impl into another?

finally figured out how to not make merges from master show up as well. 

LMK and I can get them staged.

semuxproject:
I will do another round of reviewing and then purge it.

semuxproject:
There are various small issues. I will merge and fix it

semuxproject:
Merged

Issue #324, opened 2017-12-14
Auto-generated API Documentation


cryptokat:
Currently the API documentation is manually maintained in wiki causing following problems:
- hard to track changes of each version
- need to be managed manually by project owner/collaborator 

The problems can be resolved by auto-generating API docs from source code using [Swagger](https://github.com/swagger-api/swagger-core) or a similar tool.

orogvany:
pr #449 starts to address this.  At least it provides a location for annotations to attach to.

cryptokat:
@orogvany Cool. Thanks for the info :+1:

orogvany:
[swagger.json](https://github.com/semuxproject/semux/files/1610706/swagger.txt)

Something like this. Very incomplete though.

PR #457, opened 2018-01-09
[wip] GUI: Add swagger and command console

### LINKED ISSUE #318: Add Debug Window Console
Hi, I would suggest that you add Debug Window Console for more option and make secretable the Private Key.
Some hacker too easy steal the private key, only click Wallet > Export Private Key.
You need to change with asterix code and user should click reaveal.

Thanks

orogvany:
Implement swagger annotations for auto client generation
and documentation.

Also add a console where users can call API calls without
running REST API.  We can add other commands to this
later, like dumpPrivKeys and other such common commands.

fixes #318 #324

### DIFF
--- a/config/semux.properties
+++ b/config/semux.properties
@@ -58,3 +58,10 @@ api.listenPort = 5171
 # Basic authentication
 api.username = YOUR_API_USERNAME
 api.password = YOUR_API_PASSWORD
+
+#================
+# UI
+#================
+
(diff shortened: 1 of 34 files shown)


orogvany:
Will submit in more managable chunks.

semuxproject:
No need to close the PR next time

```
git rebase -i
# go through a few iterative steps
git push -f
```

PR #460, opened 2018-01-09
Trivial decode Data String in TX Dialog

### LINKED ISSUE #455: "Trunsaction" window displays "Data" field as a byte code
####System information
Semux version: v1.0.0-rc.5
OS & Version: Windows 7 x64 (RUSSIAN localization).

####Actual behaviour
"Trunsaction" read only window on "Trunsaction" tab displays "Data" field as a byte code. "Data" value must be decoded.

####Steps to reproduce
image

My address is 0x9f0b3262b87d5350b81da69a8f79fd7893f9eff9.
Thank you.

phash:
resolves #455


semuxproject:
It's better to provide both text and hex (default in the transaction dialog) view like the send panel. Not every data can be parsed.

PR #461, opened 2018-01-10
API: Add swagger API annotations

### LINKED ISSUE #324: Auto-generated API Documentation

orogvany:
Swagger can be used for client generation
and documentation generation.

fixes #324

### DIFF
--- a/LICENSE_3RD
+++ b/LICENSE_3RD
@@ -70,6 +70,15 @@ SLF4J:
     https://www.slf4j.org/
     MIT License
 
+Swagger Annotations
+    Copyright 2016 SmartBear Software
+    https://swagger.io/
+    Apache License, Version 2.0
+
+JAX-RS API
+    Copyright (c) 2011-2012 Oracle
+    http://jax-rs-spec.java.net
+    CDDL License Version 1.1
 
 ================================================================================
 MIT License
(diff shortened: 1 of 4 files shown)


semuxproject:
We have strict policy on third party libraries:
1. Open source friendly (GPL is not)
2. Lightweight (library like guava/spring should be avoided because they're too heavy; they also depends on a lot of stuff)
3. Always code over JDK first and try not adding any dependencies (make the kernel easier to maintain and control)
4. Strictly specify the scope of dependencies (dependencies with `compile` scope will be packed and distributed, so they needs to be small).

Based on these policies, please revisit this PR

orogvany:
Ah, my mistake.  I can fix.  Only the annotations are probably required, and thats small/apache.

good catch!

orogvany:
Ok, errant libraries removed, only Apache and CDDL/GPLv2 with classpath exception.

semuxproject:
Looks good. Please update LICENSE_3RD

orogvany:
Ok.  Assumed wrongly license file was autogenerated.  
Licenses updated.

PR #472, opened 2018-01-12
GUI: Add command console

### LINKED ISSUE #318: Add Debug Window Console

orogvany:
Uses the existing REST api annotations
to create a console, as many people do not know
rest, but are familiar with the bitcoin console.

Later, more methods could be added here that are
not exposed as part of semuxAPI
like dump_priv_key or similiar.

fixes #318


orogvany:
WIP as unit tests are forthcoming.  Any feedback welcome on expected features in meantime.  Thanks!

orogvany:
@cryptokat or @semuxdev  - can remove WIP. Can be reviewed now.

cryptokat:
Getting this error from `getBlock <number>`
selection_238

orogvany:
Ah, yes.  That's the only overloaded method.  In your case, it's calling getBlock <hash> rather than number, and we're not validating the string is hexadecimal.  Let me dwell on that a few

PR #476, opened 2018-01-12
Replace Logback With Log4j2 for Better Logging Performance 

cryptokat:
Advantages of Log4j2:

- higher throughput
- lower latency
- lower memory footprint 
- fully async logger

#### References

- https://logging.apache.org/log4j/2.x/performance.html


semuxproject:
Looks perfect! 👍

PR #500, opened 2018-01-15
Add FastByteComparator

cryptokat:
The improvement speeds up byte array comparison by 15% ~ 20% according to benchmark result of `ByteComparisonPerformance.java`.

Inspired by:
- https://github.com/ethereum/ethereumj/blob/develop/ethereumj-core/src/main/java/org/ethereum/util/FastByteComparisons.java
- https://github.com/google/guava/blob/master/guava/src/com/google/common/primitives/UnsignedBytes.java

### DIFF
--- a/LICENSE_3RD
+++ b/LICENSE_3RD
@@ -79,6 +79,11 @@ JAX-RS API
     http://jax-rs-spec.java.net
     CDDL License Version 1.1
 
+Guava: Google Core Libraries for Java
+    Copyright (C) 2009 The Guava Authors
+    https://github.com/google/guava/
+    Apache License, Version 2.0
+
 ================================================================================
 MIT License
 ================================================================================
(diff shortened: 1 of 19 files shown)


semuxproject:
Was reading jdk9 Arrays#equals() implementation (need to check jdk8 too). They are using vectorized mismatch function for larger array. Not sure if this PR will make a significant difference.

cryptokat:
In JDK9 the native implementation of Array comparison is much faster:

```
# Run complete. Total time: 00:00:21

Benchmark                                       Mode  Cnt        Score       Error   Units
ByteComparisonPerformance.benchmarkArrayEqual  thrpt    5  3628437.549 ± 81239.308  ops/ms
ByteComparisonPerformance.benchmarkFastEqual   thrpt    5   780858.709 ± 15885.661  ops/ms

Process finished with exit code 0
```

PR #505, opened 2018-01-17
GUI: Add named accounts

### LINKED ISSUE #375: Add ability to label receiving addresses in the GUI.
Add ability to label receiving addresses in the GUI.

I.e:
| # | Label | Address | Available | Locked |

Transaction tab could show both label and address:
I.e:

From/To: Grandma (0x1234)

orogvany:
Adds ability to name receiving accounts.
The name is updated across all panels.
Persists with wallet.

Backwards compatibility tests are added
to support reading from old wallet formats
Default names are applied.

fixes #375


orogvany:
After this, I'll probably want to address the 'send to' to make it a dropdown/edit text, instead of clicking through to address book.

Merge address book and current accounts, for transferring between accounts.
One step at a time though.

semuxproject:
After skimming through your PR, I think we should merge the addressbook and wallet address aliases features, and upgrade the wallet.

Instead of updating the original wallet format layout, we can append a section containing all the `address_to_name` mappings at the end. Also, we can enable VLQ now, which is default ,since a new version is introduced.

At the wallet, implement a function which returns `Optional<String>` instead of giving a default name which depends on GUI (cross-dependencies). Personally, I don't the `Account #N` name any more. How about we all use `SwingUtil#shortAddress()`.

orogvany:
Yes, if we added addressbook to wallet at the end, it would make more sense to keep them separate in the wallet.

I agree that Acct# isn't ideal. Particularly when you add/remove a lot, the numbering is rather meaningless.  Ah, I see what you mean about Optional<String>.  Rather than having to handle not found, we can just make it return shortAddress if it's not in our addressbook.

orogvany:
added [wip] while I work through

- should our UI use delegateName everywhere if its a delegate?
- How to gracefully deprecate AddressBook, in favor of using the wallet

orogvany:
This looks ready to me.  Please let me know if there's any additional things needed, or anything I missed, and I'll fix ASAP.  I looked at travis, and couldn't tell why it failed.  Some timeouts on unchanged tests. :/

Couldn't figure out scrutinizer complaints.  I had messaged you about proper use of Optional checks that were failing, when it looked like they were checked appropriately. 

Thanks.

PR #526, opened 2018-01-21
API: add data field TransactionType#blockNumber

### LINKED ISSUE #525: API feature request: get_transaction, get_account_transactions - Retrieve block number and/or block hash
#### System information
Semux version: v1.0.0-4a83a2d
OS & Version: Linux/amd64

#### Expected behaviour
It would be nice if the get_transaction and get_account_transactions API calls would return the block hash and block number that the transaction was forged in.

#### Actual behaviour
These API calls only return the transaction hash (and other useful metadata).

Justification:

Being able to know the block number that a particular transaction was forged in will allow the development of software that is dependent upon the ordering of transactions. For example, validator pooling software needs to keep track of when addresses voted and unvoted.  By knowing the block number that these happened in, a timeline can be established to determine the correct payout for the validator pool.  Timestamps cannot be relied upon because of the possibility for time drift.

cryptokat:
resolve #525


semuxproject:
Note: the block number for pending transactions will be -1.

PR #528, opened 2018-01-22
Core: fix BlockchainImpl#getTransactionBlockNumber

cryptokat:
handle a special case for coinbase transactions


bjguillot:
I installed your fixed code earlier today from the develop branch and the blockNumber attribute appears to be working now for all transaction types including COINBASE.  Thanks!  (Discord user "joyteq#3002".)

PR #549, opened 2018-01-23
API: Add sign/verify messages calls

### LINKED ISSUE #509: Add ability to sign a message
It would be nice to support message signing/verification

This can just be a REST feature, or use the console.

But, it would be nice to be able to verify that someone online purporting to be a validator or someone to transact with indeed owns the address.

Note: I believe that most change this slightly from the transaction signing, so that the feature cannot be used for signing raw transaction messages for spoofing attempts

localhost:5171/sign_message?address=0x1234...&message=whatever 1/1/18
and localhost:5171/verify_message?address=0x1234..&message=whatever&signature=2k5jl12j524...

orogvany:
Use the existing signatures to construct signed messages
This allows users to verify they control a given wallet
for trust

fixes #509

### DIFF
--- a/src/main/java/org/semux/api/SemuxApiImpl.java
+++ b/src/main/java/org/semux/api/SemuxApiImpl.java
@@ -55,7 +55,6 @@ import org.semux.crypto.Key;
 import org.semux.crypto.cache.PublicKeyCache;
 import org.semux.net.NodeManager;
 import org.semux.net.filter.SemuxIpFilter;
-import org.semux.util.Bytes;
 
 public class SemuxApiImpl implements SemuxApi {
     private Kernel kernel;
--- a/src/main/java/org/semux/api/SemuxApiImpl.java
+++ b/src/main/java/org/semux/api/SemuxApiImpl.java
@@ -435,14 +435,15 @@ public class SemuxApiImpl implements SemuxApi {
             byte[] signatureAddress = Hash.h160(pubKey.getEncoded());
             byte[] addressBytes = Hex.decode0x(address);
 
+            boolean isValidSignature = true;
             if (!Arrays.equals(signatureAddress, addressBytes)) {
-                return failure("Signature does not match provided address.");
+                isValidSignature = false;
             }
             if (!Key.verify(message.getBytes(), sig)) {
-                return failure("Signature does not match message.");
+                isValidSignature = false;
             }
 
-            return new VerifyMessageResponse(true);
+            return new VerifyMessageResponse(true, isValidSignature);
         } catch (NullPointerException | IllegalArgumentException e) {
             return failure(String.format("Invalid Signature"));
         }
--- a/src/main/java/org/semux/api/SemuxApiImpl.java
+++ b/src/main/java/org/semux/api/SemuxApiImpl.java
@@ -402,15 +402,25 @@ public class SemuxApiImpl implements SemuxApi {
     @Override
     public ApiHandlerResponse signMessage(String address, String message) {
         if (address == null) {
-            return failure("Parameter `address` can't be null");
+            return failure("Parameter `address` is required");
         }
         if (message == null) {
-            return failure("Parameter `message` can't be null");
+            return failure("Parameter `message` is required");
         }
         try {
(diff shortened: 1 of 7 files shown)


src/main/java/org/semux/api/response/VerifyMessageResponse.java
```diff
@@ -0,0 +1,20 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.semux.api.ApiHandlerResponse;
+
+/**
+ */
+public class VerifyMessageResponse extends ApiHandlerResponse {
+
+    public VerifyMessageResponse(
+            @JsonProperty("success") Boolean success) {
+        super(success, null);
```
semuxproject: I thought the `success` variable only indicates if the request was successfully processed, not the result. Should we introduce another field, like `isValid`? @cryptokat

cryptokat: Yes, `success` field usually indicates whether there is an error or not. The result of message verification can be returned as `result` field conventionally.

orogvany:
Ok. Makes sense.  Its rather odd the cases where success=true but result=false though.  I guess success where params are set but result on signing success.

orogvany:
You may want to consider changing sendTransactionResponse as well
It is implemented the same ways as mine

```
 SendTransactionResponse extends ApiHandlerResponse {

    public SendTransactionResponse(
            @JsonProperty("success") Boolean success) {
        super(success, null);
    }
}
```

src/main/java/org/semux/api/response/VerifyMessageResponse.java
orogvany: See below comment.   I can change mine, but this is inline with AddNodeResponse and SendTransactionResponse.  I thought matching existing behavior was appropriate here, rather than going off the rails.  I can change the others to match if you feel they are in error as well?

cryptokat: I was thinking about introducing JSON-RPC as API version 2 such that we don't have to worry about the format of response.

semuxdev: I don't think the AddNodeResponse and VerifyMessageResponse are the same. `/verify_message` is more like `tell me if the signature is valid`, from an API user's perspective; it's more like an query rather than an operation, as in `/add_node`.

orogvany: Ok, fair enough.  I'll update to add result as boolean.

src/main/java/org/semux/api/SemuxApiImpl.java
```diff
@@ -394,6 +399,56 @@ public ApiHandlerResponse getTransactionLimits(String type) {
...
+        if (address == null) {
+            return failure("Parameter `address` can't be null");
+        }
+        if (message == null) {
+            return failure("Parameter `message` can't be null");
+        }
+        try {
+            byte[] addressBytes = Hex.decode0x(address);
+
+            Key account = kernel.getWallet().getAccount(addressBytes);
```
cryptokat: An error message like `The provided address doesn't belong to the wallet` is suggested to be returned when `account` is null.

orogvany:
Added additional validations common to other methods as well.

src/main/java/org/semux/api/SemuxApiImpl.java
```diff
@@ -394,6 +399,72 @@ public ApiHandlerResponse getTransactionLimits(String type) {
...
+            if (!Arrays.equals(signatureAddress, addressBytes)) {
+                isValidSignature = false;
+            }
+            if (!Key.verify(message.getBytes(), sig)) {
+                isValidSignature = false;
+            }
+
+            return new VerifyMessageResponse(true, isValidSignature);
+        } catch (NullPointerException | IllegalArgumentException e) {
+            return failure(String.format("Invalid Signature"));
```
semuxdev: I think we should return `result = false` instead of a failure if the signature is invalid.

orogvany: done. thanks

PR #555, opened 2018-01-24
API: Pretty print response

orogvany:
Pretty print is easier for human legibility
And does not add much extra payload size


cryptokat:
Can this be optional by adding a `pretty=true|false` query parameter?

orogvany:
Indeed it can!

cryptokat:
Thanks! LGTM except for the formatting error

orogvany:
Formatting fixed.

PR #568, opened 2018-01-24
Properly Separate Mainnet and Testnet

cryptokat:
#### Issue

As of v1.0.0 it is possible for a mainnet node to connect to a testnet node or the other way around.

#### Solution

A new mandatory capability `SEM_TESTNET` is created in place of `SEM` for testnet nodes.

### DIFF
--- a/src/main/java/org/semux/config/TestnetConfig.java
+++ b/src/main/java/org/semux/config/TestnetConfig.java
@@ -11,7 +11,7 @@ import static org.semux.net.Capability.SEM_TESTNET;
 import org.semux.Network;
 import org.semux.net.CapabilitySet;
 
-public class TestnetConfig extends AbstractConfig implements Config {
+public class TestnetConfig extends AbstractConfig {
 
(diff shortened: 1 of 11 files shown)


src/main/java/org/semux/config/TestnetConfig.java
```diff
@@ -6,11 +6,19 @@
  */
 package org.semux.config;
 
+import static org.semux.net.Capability.SEM_TESTNET;
+
 import org.semux.Network;
+import org.semux.net.CapabilitySet;
 
-public class TestnetConfig extends AbstractConfig {
+public class TestnetConfig extends AbstractConfig implements Config {
```
orogvany: AbstractConfig already implements Config

cryptokat: fixed

semuxdev:
I was hoping that we have a `networkId` in the `Peer` data structure.

Luckily, we have the capability set! 👍

PR #576, opened 2018-01-24
GUI: Import wallet imports addressbook too

### LINKED ISSUE #565: Recovery procedure doesn't restore a saved address book
#### System information
Semux/v1.0.0-4a83a2d/Windows/amd64

OS & Version: Windows 7x64

#### Expected behaviour
Wallet backup contains an address book. "Restore wallet" procedure recoveres the saved address book.

#### Actual behaviour
A wallet file stores an address book, but recovery procedure doesn't restore a saved address book.

Restore a wallet with an addressbook from a backup. The address book is empty (isn't recovered).

Bounty address is 0x9f0b3262b87d5350b81da69a8f79fd7893f9eff9

orogvany:
Import the new wallets address book.
Chose to not override any existing aliases, but could
go either way on that one.  There is little chance of overlap.

Prefer to keep existing wallet as source of truth if conflicts arise

fixes #565

### DIFF
--- a/src/main/java/org/semux/core/Wallet.java
+++ b/src/main/java/org/semux/core/Wallet.java
@@ -515,4 +515,31 @@ public class Wallet {
             throw new WalletLockedException();
         }
     }
+
+    /**
+     * Adds the addresses and aliases from another wallet
+     *
+     * @param w
+     * @return
+     */
+    public int addWallet(Wallet w) {
+        requireUnlocked();
+
+        // import addresses
+        int numImportedAddresses = addAccounts(w.getAccounts());
+
+        // add address book entries
+        Map<ByteArray, String> importedAliases = w.getAddressAliases();
+        for (Map.Entry<ByteArray, String> alias : importedAliases.entrySet()) {
+            byte[] address = alias.getKey().getData();
+
+            // don't override existing wallet's aliases.
+            if (!getAddressAlias(address).isPresent()) {
+                setAddressAlias(address, alias.getValue());
+            }
+        }
+
+        flush();
+        return numImportedAddresses;
+    }
 }
--- a/src/main/java/org/semux/core/Wallet.java
+++ b/src/main/java/org/semux/core/Wallet.java
@@ -24,6 +24,7 @@ import org.semux.crypto.Aes;
 import org.semux.crypto.CryptoException;
 import org.semux.crypto.Hash;
 import org.semux.crypto.Key;
+import org.semux.message.CliMessages;
 import org.semux.util.ByteArray;
 import org.semux.util.Bytes;
 import org.semux.util.IOUtil;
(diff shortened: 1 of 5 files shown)


src/main/java/org/semux/gui/MenuBar.java
```diff
@@ -182,6 +184,17 @@ protected void recoverAccounts() {
 
                 Wallet wallet = gui.getKernel().getWallet();
                 int n = wallet.addAccounts(w.getAccounts());
+                // add address book entries
+                Map<ByteArray, String> importedAliases = w.getAddressAliases();
+                for (Map.Entry<ByteArray, String> alias : importedAliases.entrySet()) {
+                    byte[] address = alias.getKey().getData();
+
+                    // don't override existing wallet's aliases.
+                    if (wallet.getAddressAlias(address) == null) {
```
cryptokat: This needs to be changed to `!wallet.getAddressAlias(address).isPresent()`

orogvany:
I'll extract the wallet import code, so both places can use same importing.  Was not aware that it was done separately 2 different places

src/main/java/org/semux/gui/MenuBar.java
orogvany: done, thanks

src/main/java/org/semux/gui/WelcomeFrame.java
```diff
@@ -283,7 +283,7 @@ protected void ok() {
                 JOptionPane.showMessageDialog(this, GuiMessages.get("NoAccountFound"));
             } else {
                 if (wallet.unlock(password)
-                        && wallet.addAccounts(w.getAccounts()) > 0
+                        && wallet.addWallet(w) > 0
                         && wallet.flush()) {
```
cryptokat: The wallet flushes 2 times here.

src/main/java/org/semux/core/Wallet.java
```diff
@@ -515,4 +515,31 @@ private void requireUnlocked() throws WalletLockedException {
...
+        for (Map.Entry<ByteArray, String> alias : importedAliases.entrySet()) {
+            byte[] address = alias.getKey().getData();
+
+            // don't override existing wallet's aliases.
+            if (!getAddressAlias(address).isPresent()) {
+                setAddressAlias(address, alias.getValue());
+            }
+        }
+
+        flush();
```
cryptokat: It would provide better UX if we report an error when the wallet is failed to save on disk.

orogvany: done.

src/main/java/org/semux/gui/WelcomeFrame.java
orogvany: You're right.  The addWallet method should not call flush() for consistency with other wallet modifying methods.  This has been corrected.

orogvany: Not sure how to resolve this change request?

PR #582, opened 2018-01-26
[WIP] API: Do not allow certain API calls on locked wallet

### LINKED ISSUE #579: The api still work after i locked wallet
Hi there,

#### System information

Semux version: v1.0.0-rc.1

OS & Version:Windows 10

#### Expected behaviour

the api is locked too， and can not do transaction

#### Actual behaviour

the api still work

1. lock the wallet
2. check api status
#### Backtrace

And this is my address for the bounty:
0xf6e80170fc82e49b8a23a178a6c9e84b4d5fcb2f

orogvany:
Allow informational requests that do not contain personal
account information, but do not allow account modifying methods
to be called with locked wallet.

fixes #579


semuxdev:
I think we should category our commands instead of manually adding these checks.

For instance, `query`, `operation`, `wallet` and API provider should be able to choose which category to open

orogvany:
If we were to go that route, I think I might prefer to group them at URL path level as well
http://host/wallet/send_transaction /query/get_block etc

to make it easier for other folks to use filtering rules as to what can be exposed, etc

semuxdev:
I like your idea, but my main concerns is it breaks compatibility.

orogvany:
You could support the old APIs as well, as redirects.

I guess we'll also need to figure API schema versioning at some point too, as there will come a time when new 'required' fields are required, etc.

orogvany:
Ok, for now, let's leave the API paths alone, and just add groupings to methods.

if grouping != query, we have a central place that checks locked

orogvany:
I initially implemented this with Annotations on the methods, but I've since extracted out interfaces for the various APIs.

For checking locked, we can do this in a central location using a proxy, if we want to leave the API handler untouched.

Or we can just make the APIhandler aware of the 3 different API interfaces.  Thoughts?

Issue #534, opened 2018-01-22
Optimize Wallet for Large Number of Accounts


cryptokat:
#### Expected behaviour

The wallet should be able to handle large number of accounts without significantly impacting performance in case of online crypto exchanges.

#### Actual behaviour

The wallet loads all accounts into memory space which might eventually result in OOM crash.

testzcrypto:
Exchanges can use one account and text in data as additional identifier.

orogvany:
If you have a large number of accounts, its noted that dropdowns like validator window make it *very* difficult to select lower ones, as the refresh happens quite a bit and breaks scrollling

gpip:
Exchanges certainly won't be using this method for creating accounts. Maybe some bad ones, but beyond those it's more common to have a bip32-like derivation. None of the generated addresses would be stored in semux's wallet.data. Deposits would be checked against pending transactions + checking transactions included in blocks.

gpip:
@testzcrypto usually that's only done with coins that have various issues with handling many accounts. For xrp and xlm there's a cost associated with account creation, so it becomes prohibitive to create millions of accounts. For monero it's somewhat hard and slow to manage many accounts so integrated addresses are used.

cryptokat:
Hi @gpip 
Thank you for your feedback.
As we discussed in #621, this was removed from the todo list for exchange listing.
Please see https://github.com/semuxproject/semux/projects/3 for the latest todo list.

PR #601, opened 2018-01-28
Core: Wallet lookups do not iterate list

orogvany:
Resolve TODOs around wallet lookups.
Use LinkedHashMap to maintain order.

We're using a synchronized collection
yet usages look to all be protected with
synchronized.  It might be better to not use
synchronized wrapper, after double checking
synchronization.

Would like to discuss, and possibly add another
test before merging.

### DIFF
--- a/src/main/java/org/semux/core/Wallet.java
+++ b/src/main/java/org/semux/core/Wallet.java
@@ -14,6 +14,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


orogvany:
Huh, I net removed code, but code coverage went down.  Very strange.

src/main/java/org/semux/core/Wallet.java
```diff
@@ -314,7 +317,9 @@ public void setAccounts(List<Key> list) throws WalletLockedException {
     public Key getAccount(int idx) throws WalletLockedException {
         requireUnlocked();
 
-        return accounts.get(idx);
+        synchronized (accounts) {
+            return getAccounts().get(idx);
```
semuxdev: Just wanted to add a note here. One iteration over all keys are required here.

orogvany: Yes, it is a tradeoff.  As a list, that method is cheap, but the other 4-5 places are O(n) time.  With a map, the other methods are constant time, and this one is O(n) time.

As there's never likely to be over 100 accounts, it's all pretty trivial.  I could introduce another collection to keep track of index as well, but this seemed like an ok place to start.

cryptokat:
related: #534

Issue #578, opened 2018-01-25
The Transaction panel should show more Decimals


likaigithub:
Hi there,

#### System information

Semux version: v1.0.0-rc.1

OS & Version:Windows 10

#### Expected behaviour

show 0.000000001 SEM 

#### Actual behaviour

show 0.000 SEM

Description, screenshot allowed

#### Steps to reproduce the behaviour

I voted a validator with 0.000000001 SEM，and you can check here
[http://semuxchain.info/transaction/0xf3cb00c915d700251ad596120ce1a12b4e8bcf1eb428ca5eccf49188726f3783](http://semuxchain.info/transaction/0xf3cb00c915d700251ad596120ce1a12b4e8bcf1eb428ca5eccf49188726f3783)

then i check transaction panel  ,the amount is 0.000 SEM
![_ 24yz p y_aj dlm1axm](https://user-images.githubusercontent.com/9426498/35372332-5252cc9c-01d4-11e8-81f6-d49b1a8e3b50.png)

#### Backtrace

And this is my address for the bounty:
0xf6e80170fc82e49b8a23a178a6c9e84b4d5fcb2f

orogvany:
This can be closed.

cryptokat:
duplicate: #608

Issue #569, opened 2018-01-24
Delegates panel should show decimals for low values


orogvany:
If I vote say 0.6 SEM for a validator for whatever reason
the panel will say votes = 1

If i then go and try to unvote 1, it gives an error

similarly, if I vote 19.99, it will show 20, and I will be unable to unvote properly.

Also, this can lead to fractional SEM remaining locked, despite delegate votes all showing 0
It should always round down, or if <1, show the fractional remains
Or always show fractional numbers

cryptokat:
closed by #681

PR #602, opened 2018-01-28
GUI: Render to highest precision

orogvany:
For delegates panel, we show 0 for < 1 sem
As we allow fractional SEM votes, we should show
to same precison as elsewhere.

Also only show precision to just number available.
So we will render 3 instead of 3.000 and 3.2 instead
of 3.200.  This can allow us to choose to render more
decimals if we want without cluttering UI if user only
uses full SEM.

this also helps where folks say they believe that other
validators with 'same' number of votes took their spot
when its likely just a fractional vote that pushed them over

### DIFF
--- a/src/main/java/org/semux/gui/SwingUtil.java
+++ b/src/main/java/org/semux/gui/SwingUtil.java
@@ -311,6 +311,7 @@ public class SwingUtil {
         NumberFormat format = NumberFormat.getInstance();
         format.setMinimumFractionDigits(decimals);
         format.setMaximumFractionDigits(decimals);
+        format.setMinimumFractionDigits(0);
 
         return format.format(number);
     }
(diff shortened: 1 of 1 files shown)


cryptokat:
related: #578, #569

src/main/java/org/semux/gui/SwingUtil.java
```diff
@@ -311,6 +311,7 @@ public static String formatNumber(Number number, int decimals) {
         NumberFormat format = NumberFormat.getInstance();
         format.setMinimumFractionDigits(decimals);
         format.setMaximumFractionDigits(decimals);
+        format.setMinimumFractionDigits(0);
```
semuxdev: set min fraction digits twice

orogvany: thanks

PR #605, opened 2018-01-29
Tools: upgrade Jackson to 2.9.4

cryptokat:
to avoid [CVE-2017-17485](https://nvd.nist.gov/vuln/detail/CVE-2017-17485)


orogvany:
Seems requires spring to exploit. But good to keep current

PR #606, opened 2018-01-29
GUI: Update to new logo

orogvany:
Update image/icon to new semux logo.


orogvany:
Noticed welcome pane was distorted, updated

PR #625, opened 2018-01-31
Initialize a Validator-Activated Softfork UNIFORM_DISTRIBUTION

### LINKED ISSUE #620: Incorrect validator number distribution function
Hi!

#### Expected behaviour

The Semux client should choose validator to forge a block based on uniform distribution (so any validator on a long period of time should forge so many blocks as other).

#### Actual behaviour

Validators with numbes 1-56 has an advantage over others. The difference is about 1.5x
For example, if validator with number 1-56 will forge 150 blocks, others will forge only 100 blocks each.

I found this bug when my validator, which is constantly at the bottom of the table, forged significantly fewer blocks than it should be by calculation.

The problem is: 

	validators.get((Hash.h256(key)[0] & 0xff) % validators.size());

h256 is uniform on 0->255, but after %100 we get not uniform distibution. 
Cause first 200 give a uniform distribution, but the additional 55 corrupt everything.

For a visual demonstration, I plotted a graph showing the problem. 
The series2 is result of working h256 function, the series1 is result of h256 % 100.
distribution

#### Solution

I think you should use something like this: 

	long seed = new BigInteger(key).longValue();
	int number = new Random(seed).nextDouble() * 100;

This will give a uniform distribution and it's gives the same values for the same sid (height and view).

But I don't know this java Random implementation is finish or not. I.e., whether it can changed in the future or not.
Perhaps it's safer to use its implementation:

	public static int random100(long seed) {
		long oldSeed, nextSeed;
		AtomicLong s = new AtomicLong(seed);
		do {
			oldSeed = s.get();
			nextSeed = (oldSeed * 0x5DEECE66DL + 0xBL) & ((1L << 48) - 1);
		} while (!s.compareAndSet(oldSeed, nextSeed));
		int nextInt = (int)(nextSeed >>> (48 - 24));
		return (int) (nextInt / (float)(1 << 24) * 100);
	}

Returns [0-100).
It's based on default Java implemantion. 
I tested this it's give a uniform distrubion as expected.
The deviation of distribution is about 0.2%.
But I did not test this solution very long. Probably I did not take something into account. 
And maybe there is another simpler method of solving the problem.
And as I understand we need hardwork for it and you should save old function before hardfork block.

#### Test
		
	public static void main(String[] args) {
		byte []key;
		Random random = new Random();
		
		HashMap<Integer, AtomicInteger> map100 = new HashMap<>();
		HashMap<Integer, AtomicInteger> mapPRNG = new HashMap<>();
		HashMap<Integer, AtomicInteger> map256 = new HashMap<>();
		
		for (int i = 0; i < 256; i++) {
			map256.put(i, new AtomicInteger(0));
			if(i < 100) {
				map100.put(i, new AtomicInteger(0));
				mapPRNG.put(i, new AtomicInteger(0));
			}
		}

		int view, v256, v100, vPNRG;
		long seed;
		for (int i = 0; i < 1_000_000; i++) {
			view = random.nextDouble() < 0.05 ? 1: 0; //about 5%
			key = Bytes.merge(Bytes.of(i), Bytes.of(view));
			v256 = Hash.h256(key)[0] & 0xff;
			v100 = v256 % 100;

			seed = new BigInteger(Bytes.merge(Bytes.of(i), Bytes.of(view))).longValue();
			//vPNRG = (int)(new Random(seed).nextDouble() * 100);
			vPNRG = random100(seed);

			map100.get(v100).incrementAndGet();
			map256.get(v256).incrementAndGet();
			mapPRNG.get(vPNRG).incrementAndGet();
		}
		System.out.println("Number\th256\t%100\tPRNG");
		for (int i = 0; i < 256; i++) {
			System.out.println(i + "\t" + map256.get(i).intValue() +
					"\t" + (map100.containsKey(i)? map100.get(i).intValue(): 0) +
					"\t" + (mapPRNG.containsKey(i)? mapPRNG.get(i).intValue(): 0)
			);
		}
	}

	public static int random100(long seed) {
		long oldSeed, nextSeed;
		AtomicLong s = new AtomicLong(seed);
		do {
			oldSeed = s.get();
			nextSeed = (oldSeed * 0x5DEECE66DL + 0xBL) & ((1L << 48) - 1);
		} while (!s.compareAndSet(oldSeed, nextSeed));
		int nextInt = (int)(nextSeed >>> (48 - 24));
		return (int) (nextInt / (float)(1 << 24) * 100);
	}

cryptokat:
To resolve #620

- [x] Initialize a validator-activated softfork `UNIFORM_DISTRIBUTION` based on `BlockHeaderData#forkNumber`. The softfork will be activated when there are 75% (1500) of past 2000 blocks signaling the fork.
- tests
  - [x] 75% rule works
  - [x] a legacy node should be able to accept a block made with the new rule
  - [x] a legacy validator node should be able to accept a block made with the new rule
- [x] optimize block lookup for fork activation from O(m) to O(1)
- [x] add a deadline height to expire a soft fork
- [x] add a database migration to fix validator stats for previous versions of wallet
- [x] comments & cleanup
- [x] code review

This implementation is inspired by [BIP-34](https://github.com/bitcoin/bips/blob/master/bip-0034.mediawiki).


src/main/java/org/semux/consensus/Fork.java
```diff
@@ -0,0 +1,40 @@
...
+ * LICENSE or https://opensource.org/licenses/mit-license.php
+ */
+package org.semux.consensus;
+
+public class Fork {
+
+    public static final Fork UNIFORM_DISTRIBUTION = new Fork(190, 200, 0);
+
+    /**
+     * The number of blocks which are required to contain the fork-bit inside ofg
```
orogvany: typo 'ofg'

src/main/java/org/semux/consensus/SemuxBft.java
```diff
@@ -308,9 +314,9 @@ protected void enterPropose() {
         }
 
         logger.info("Entered propose: height = {}, view = {}, primary = {}, # connected validators = 1 + {}", height,
-                view, isPrimary(), activeValidators.size());
+                view, isPrimary(uniformDistributionActivated), activeValidators.size());
```
orogvany: uniformDistributionActivated is class level, so shouldn't need to be passed as param here.

src/main/java/org/semux/consensus/SemuxBft.java
```diff
@@ -652,6 +659,12 @@ public void onMessage(Channel channel, Message msg) {
         }
     }
 
+    protected void activateForks() {
```
orogvany: Would consider adding TODO or note otherwise to remove this after fork has been successfully activated so as not to clutter future releases, or restarts of testnet.

src/main/java/org/semux/core/BlockHeader.java
```diff
@@ -117,6 +118,10 @@ public boolean validate() {
                 && Arrays.equals(Hash.h256(encoded), hash);
     }
 
+    public boolean hasDataBit(int bit) {
+        return BitSet.valueOf(getData()).get(bit);
```
orogvany: This will limit future uses of data's fucntionality to just using it for forking.  Maybe some more structured object would be appropriate here.

cryptokat: @orogvany Thanks for the review! 
I wasn't able to find a more appropriate data field to store the signal other than `BlockHeader#data` without breaking backward-compatibility. The data field is totally unused at the moment.
Any suggestion would be appreciated.

cryptokat: @orogvany Thanks for the review!
Would something like this be more appropriate than a BitSet:

```
class BlockHeaderData {
   // 8 bytes: we will be able to introduce at most 8 * 8 = 64 forks in the future
   public final BitSet forkBits;

   // 24 bytes: reserved for other usage
   public final byte[] reserved;
}
```

orogvany: Yes, that's more of what I was thinking.

I do imagine that once a fork has occurred, that the code to migrate to it could be removed in the future, or I guess more carefully if more than 64 hardforks is required, code could be implemented to roll back to 000000...  Or instead of bitset, a 'fork number' could be used, so you would only need 1 byte to handle 255 forks. 

I don't think a bitset is necessarily required, as there should only be one fork at a time.

cryptokat: improved `BlockHeaderData` without a BitSet
```
class BlockHeaderData {
   // 2 bytes: an increment number as more forks are introduced
   public final short forkNumber;

   // 30 bytes: reserved for other usage
   public final byte[] reserved;
}
```

orogvany: 👍 Leaves plenty of room for other future use down the road.

semuxdev:
Should we have a fixed fork point rather than fork whenever the preset condition is met?  IMO it may cause trouble for payment system, exchanges, block explorer to react to the pending forks.

cryptokat:
@semuxdev 
The main idea of this implementation is to apply a new consensus rule without breaking backward-compatibility. It should be considered as a bug if this implementation introduces any incompatibility with previous versions of Semux wallet.

Master-Bruce:
1 - Would going ahead with a soft fork not create an opportunity for some to exploit the system by submitting blocks that are only valid for when using the current (broken) random function? 

It seems to me it would be safer to have a fixed fork after which the correct uniform random function guards the chain.

orogvany:
@Master-Bruce It wouldn't allow exploit, but validators could just not update and hold this change back.  It seems best to allow validators to agree to update first, and if concensus is not triggered, then a more aggressive block triggering could be signaled.  Mandatory upgrade isn't signaled for a long while yet (~150 days).

Aggressive triggering will mean any validators that did not update will lose forgings and network will weaken. It's preferable to have network agree a time to update rather than being forced to.

src/main/java/org/semux/core/BlockchainImpl.java
```diff
@@ -312,7 +316,7 @@ public synchronized void addBlock(Block block) {
 
             // [5] update validator statistics
             List<String> validators = getValidators();
-            String primary = config.getPrimaryValidator(validators, number, 0);
+            String primary = config.getPrimaryValidator(validators, number, 0, uniformDistributionActivated);
             adjustValidatorStats(block.getCoinbase(), StatsType.FORGED, 1);
             if (primary.equals(Hex.encode(block.getCoinbase()))) {
```
cryptokat: Incompatibility: The validator stats will start recording wrong data after fork activation for previous versions of Semux wallet.

cryptokat: Luckily we are not yet using validator stats as a part of the consensus. A database migration could solve the issue.

AlexAltertrader:
150 days of incorrect distribution of reward is monstrously a lot.

orogvany:
Sorry @AlexAltertrader that you misread what I said.  More aggressive strategy does not include the already-in-place mandatory upgrade in 150 days, hence _more_ aggressive, and the entire following paragraphs. :D

AlexAltertrader:
@orogvany Now I understood what you wanted to say.
Give a chance to the validators to do good, and if they do not want to do it, do it badly in 150 days :)

src/main/java/org/semux/config/AbstractConfig.java
```diff
@@ -148,9 +150,25 @@ public int getNumberOfValidators(long number) {
     }
 
     @Override
-    public String getPrimaryValidator(List<String> validators, long height, int view) {
-        byte[] key = Bytes.merge(Bytes.of(height), Bytes.of(view));
-        return validators.get((Hash.h256(key)[0] & 0xff) % validators.size());
+    public String getPrimaryValidator(List<String> validators, long height, int view, boolean uniformDist) {
+        if (uniformDist) {
```
orogvany: I would still recommend adding TODO or otherwise to indicate what our long term strategy for what we do after successful branch will be.

src/main/java/org/semux/core/BlockchainImpl.java
orogvany: including validator upgrade for stats would be preferrable.  I think you're on the correct tract here though

src/main/java/org/semux/config/AbstractConfig.java
cryptokat: added

cryptokat:
Testing on the testnet with a single validator node.
Coinbase = `0xce52d798aedb4e39327b0fef01cf18c0aaa1a4fc`

cryptokat:
On the testnet:
There are currently 7 validators: 6 develop branch + 1 soft fork enabled.
The number of activated blocks stopped growing at 317 ~ 320 in the past 2000 blocks.
```
13:48:51.767 DEBUG    BlockchainImpl   Fork activation of UNIFORM_DISTRIBUTION: 318 / 1900 (activated = false) in the past 2000 blocks
```

src/main/java/org/semux/config/AbstractConfig.java
```diff
@@ -148,9 +155,26 @@ public int getNumberOfValidators(long number) {
...
+    public String getPrimaryValidator(List<String> validators, long height, int view, boolean uniformDist) {
+        // TODO: add a checkpoint once UNIFORM_DISTRIBUTION is fully activated
+        if (uniformDist) {
+            long seed = new BigInteger(Bytes.merge(Bytes.of(height), Bytes.of(view))).longValue();
+            return validators.get(deterministicRandom(seed, validators.size()));
+        } else {
+            byte[] key = Bytes.merge(Bytes.of(height), Bytes.of(view));
+            return validators.get((Hash.h256(key)[0] & 0xff) % validators.size());
+        }
+    }
```
semuxdev: If `height > 0xffffffff`, the primary validator selection will be broken. It takes long though :)

src/main/java/org/semux/core/BlockchainImpl.java
```diff
@@ -60,12 +63,15 @@
...
 
+    protected static final int DATABASE_VERSION = 1;
+
     protected static final byte TYPE_LATEST_BLOCK_NUMBER = 0;
     protected static final byte TYPE_VALIDATORS = 1;
     protected static final byte TYPE_VALIDATOR_STATS = 2;
     protected static final byte TYPE_BLOCK_HASH = 3;
     protected static final byte TYPE_TRANSACTION_HASH = 4;
     protected static final byte TYPE_ACCOUNT_TRANSACTION = 5;
+    protected static final byte TYPE_DATABASE_VERSION = Integer.valueOf(999).byteValue();
```
semuxdev: `0xE7` looks odd to me.

src/main/java/org/semux/core/BlockchainImpl.java
```diff
@@ -478,6 +501,29 @@ protected void addTransactionToAccount(Transaction tx, byte[] address) {
...
     }
 
+    /**
+     * Returns the version of current database.
+     *
+     * @return
+     */
+    protected int getDatabaseVersion() {
+        byte[] versionBytes = indexDB.get(getDatabaseVersionKey());
+        if (versionBytes.length == 0) {
```
semuxdev: `versionBytes == null`

semuxdev:
Also, is it possible that the fork points shift when replaying the whole history, since old block header data will be re-interpreted as more fork numbers are introduced?

src/main/java/org/semux/core/BlockHeaderData.java
```diff
@@ -0,0 +1,59 @@
...
+    public BlockHeaderData(short forkNumber) {
+        this.forkNumber = forkNumber;
+    }
+
+    private BlockHeaderData() {
+        this.forkNumber = null;
+    }
+
+    public boolean forkActivated(ValidatorActivatedFork fork) {
+        return this.forkNumber != null && this.forkNumber >= fork.number;
```
cryptokat: Bug: `this.forkNumber >= fork.number`
This condition can shift the activation point of a fork when there are more than 1 fork to be activated.

cryptokat:
I agree with @semuxdev on adding a deadline to each fork.
What would be the appropriate deadline for this fork?
For example we give it 1 month from now, the height of deadline would be 130000.

src/main/java/org/semux/config/AbstractConfig.java
```diff
@@ -152,9 +159,16 @@ public int getNumberOfValidators(long number) {
     }
 
     @Override
-    public String getPrimaryValidator(List<String> validators, long height, int view) {
-        byte[] key = Bytes.merge(Bytes.of(height), Bytes.of(view));
-        return validators.get((Hash.h256(key)[0] & 0xff) % validators.size());
+    public String getPrimaryValidator(List<String> validators, long height, int view, boolean uniformDist) {
+        // TODO: add a checkpoint once UNIFORM_DISTRIBUTION is fully activated
+        if (uniformDist) {
+            BigInteger seed = BigInteger.valueOf(height).add(BigIntegerUtil.random(BigInteger.valueOf(view)));
```
orogvany: This is problematic I think.

On upping view (which is done when a validator misses a block), it will *always* resolve to the validator that will forge3 blocks from now

so getPrimaryValidator(...., n, 1) == getPrimaryValidator(..., n+3,0)

src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -452,7 +452,6 @@ protected boolean applyBlock(Block block, AccountState asSnapshot, DelegateState
             writeLock.unlock();
         }
 
-        current.set(block.getNumber() + 1);
```
orogvany: If you remove this, doesn't current never get set?  Or where was this moved to?

src/main/java/org/semux/util/BigIntegerUtil.java
```diff
@@ -117,6 +117,20 @@ public static BigInteger min(BigInteger v1, BigInteger v2) {
...
     }
 
+    /**
+     * Fast pseudo random generator based on LCG Algorithm. Credits to:
+     * https://software.intel.com/en-us/articles/fast-random-number-generator-on-the-intel-pentiumr-4-processor
+     *
+     * @param seed
+     * @return
+     */
+    public static BigInteger random(BigInteger seed) {
```
orogvany: In general, this random method looks to lack... randomness.  I would advise if we're going to claim randomness to use a reliably random function, else move back to round robin.

eg.
3
6
9
12
16
19
22
25
29
32
35
39
42
45
48
52
55

cryptokat: This is implemented based on LCG algorithm which is also being used in the native Random implementation of JVM. Check https://software.intel.com/en-us/articles/fast-random-number-generator-on-the-intel-pentiumr-4-processor

orogvany: That is not what's being done on the JVM.
https://docs.oracle.com/javase/7/docs/api/java/util/Random.html

I dunno.  If using JVM's Random, i don't get predictably incrementing numbers, or repeating results.  This one looks problematic and very low entropy...

orogvany: Alright, well, when used with the newer code, it appears that it's reasonably ok distribution.

I'd make sure to watch out for any other code attempting to use that random function, as I don't believe it qualifies as a random function, but for the purposes here, it's fine.

cryptokat: Fixed in https://github.com/semuxproject/semux/pull/625/commits/ef15a4ab65519615c352285ccc97c7566d39413d

The new series of result of `BigIntegerUtil#random` with an incremental seed from 0 ~ 99
```
87
84
94
91
74
71
81
78
61
58
68
65
48
45
55
51
40
36
46
43
27
23
33
30
13
10
20
17
0
97
7
4
83
80
89
86
70
67
76
73
57
53
63
60
44
40
50
47
35
32
42
38
22
19
29
25
9
6
16
12
96
93
2
99
78
75
85
82
65
62
72
69
52
49
59
56
39
36
46
42
31
27
37
34
18
14
24
21
5
1
11
8
91
88
98
95
74
71
80
77
```

orogvany: Cool, thanks!

Rest looks good to me

src/main/java/org/semux/config/AbstractConfig.java
```diff
@@ -162,7 +162,10 @@ public int getNumberOfValidators(long number) {
     public String getPrimaryValidator(List<String> validators, long height, int view, boolean uniformDist) {
         // TODO: add a checkpoint once UNIFORM_DISTRIBUTION is fully activated
         if (uniformDist) {
-            return validators.get(getUniformDistPrimaryValidatorNumber(validators.size(), height, view));
+            return validators.get(
+                    view == 0
+                            ? (int) (height % (long) validators.size())
```
orogvany: You may want to handle the case where height % validators.size() == random number for view 1
but it may not be a big deal.

Otherwise looks good.

cryptokat: This actually happens quite frequently. The chance is around 1% in the first 100K blocks.

cryptokat: Fixed in https://github.com/semuxproject/semux/pull/625/commits/5ffaec1e641f98ea083b8fa2837d67040ccdd2b7
The chance of consecutive forgers should now become 0%.

pom.xml
```diff
@@ -867,6 +868,18 @@
             <version>2.6</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.rauschig</groupId>
```
orogvany: might want to add these to the maven-enforcer-plugin.  It looks like we usually add test jars as well.

src/main/java/org/semux/config/AbstractConfig.java
```diff
@@ -156,9 +163,40 @@ public int getNumberOfValidators(long number) {
...
+        // TODO: add a checkpoint once UNIFORM_DISTRIBUTION is fully activated
+        if (uniformDist) {
+            return validators.get(getUniformDistPrimaryValidatorNumber(validators.size(), height, view));
+        } else {
+            byte[] key = Bytes.merge(Bytes.of(height), Bytes.of(view));
+            return validators.get((Hash.h256(key)[0] & 0xff) % validators.size());
+        }
+    }
+
+    public int getUniformDistPrimaryValidatorNumber(int size, long height, long view) {
```
orogvany: Probably should not be scoped public, as this is a temporary construct.

src/main/java/org/semux/config/AbstractConfig.java
```diff
@@ -156,9 +163,40 @@ public int getNumberOfValidators(long number) {
...
+        // use round-robin for view 0
+        if (view == 0) {
+            return (int) (height % (long) size);
+        }
+
+        // here we ensure there will never be consecutive block forgers after view
+        // change
```
orogvany: weird line break here.  line 185 is longer.  /pedantic off

src/main/java/org/semux/consensus/SemuxBft.java
```diff
@@ -653,6 +663,20 @@ public void onMessage(Channel channel, Message msg) {
         }
     }
 
+    /**
+     * Attempt to activate pending forks at current height.
+     */
+    private void activateForks() {
+        if (config.forkUniformDistributionEnabled()
+                && !activatedForks.containsKey(UNIFORM_DISTRIBUTION)
+                && height <= UNIFORM_DISTRIBUTION.activationDeadline
```
orogvany: Isn't the idea that activationDeadline should automatically enforce?
Or is it when we give up on trying?  Seems like once we hit activationDeadline, the activatedForks will unactivate itself?

I guess, what's the mechanism for this to keep running after activationDeadline has past?  Are we just forcing ourselves to update again by then to remove this scaffolding?

src/main/java/org/semux/core/BlockchainImpl.java
```diff
@@ -105,25 +131,60 @@ public BlockchainImpl(Config config, DbFactory factory) {
...
+    }
+
+    /**
+     * Upgrade this database from version 0 to version 1.
+     *
+     * @param dbFactory
+     */
+    private void upgradeDb0(DbFactory dbFactory) {
+        // run the migration
+        new MigrationBlockDbVersion001().migrate(config, dbFactory);
```
orogvany: Static method, rather than new X().doStuff() ?

orogvany:
So.  Want to arrange for testing on testnet?

src/main/java/org/semux/consensus/SemuxBft.java
cryptokat: The fork will never be activated before the 75% threshold is reached. Once the `activationDeadline` is reached, and the fork is still not activated, validators will stop signaling `UNIFORM_DISTRIBUTION` in block header. Then we will have to decide on if we should cancel or postpone the fork.

src/main/java/org/semux/core/BlockchainImpl.java
cryptokat: Here I am trying to borrow the concept of database migration classes from web application development. It's very common that web developers create a new class for each schema change. The benefit of this approach is providing a standardized way for developers to implement schema changes. For example: https://flywaydb.org/getstarted/java

src/main/java/org/semux/config/AbstractConfig.java
cryptokat: Yeah it's kind of weird. That's done by the formatter though.

cryptokat:
@orogvany I just installed a test build of this branch on my testnet validator. The validator is: https://testnet.semux.info/explorer/account/0xce52d798aedb4e39327b0fef01cf18c0aaa1a4fc

src/main/java/org/semux/core/BlockHeaderData.java
```diff
@@ -0,0 +1,147 @@
...
+        public ForkSignal(byte[] bytes) {
+            bitSet = BitSet.valueOf(bytes);
+        }
+
+        boolean signalingFork(ValidatorActivatedFork fork) {
+            return bitSet.get(fork.number);
+        }
+
+        byte[] toBytes() {
+            return bitSet.toByteArray();
```
semuxdev: This returns byte array of variable length, rather than a fixed 4-byte array as documented before. Also, is `BitSet` encoding language independent?

cryptokat: fixed by https://github.com/semuxproject/semux/pull/625/commits/8f217ee29b5d0c11eb995c930390c877e537fbe3

src/main/java/org/semux/consensus/SemuxBft.java
```diff
@@ -761,6 +792,27 @@ protected Block proposeBlock() {
...
+    private boolean signalingUniformDistribution() {
+        // do not signal the fork if this user decides not to participate in the fork
+        if (!config.forkUniformDistributionEnabled()) {
+            return false;
+        }
+
+        // continue signaling after fork activation
+        if (activatedForks.containsKey(UNIFORM_DISTRIBUTION)) {
+            return true;
+        }
```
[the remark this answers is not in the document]
cryptokat: In fact it doesn't make sense to continue signaling after activation, considering that there will be only 8 pending forks allowed. Fixed in https://github.com/semuxproject/semux/commit/8f217ee29b5d0c11eb995c930390c877e537fbe3

PR #642, opened 2018-02-04
JavaFX version of wallet

Vjeredy:
These are JavaFX FXML projects: Wallet application and Preloader (which at this point only shows logo while Wallet loading). In case you have troubles with building them, I also added my Netbeans properties of projects, delete if don't need.
I've hardcoded some data examples in WalletUpdater for Home and Receive scenes so you can better understand how it works.
There isn't much to tell if you are familiar with JavaFX, but even if you don't, it is quite easy to understand, just have to look in the code.
Guess this approach can solve lots of your problems with outdated GUI libraries and keep code healthier overall.


Vjeredy:
Forgot to include controlsfx library for Wallet application:
[controlsfx-8.40.14.zip](https://github.com/semuxproject/semux/files/1692565/controlsfx-8.40.14.zip)

src/main/javafx/semux/Preloader/nbproject/genfiles.properties
```diff
@@ -0,0 +1,8 @@
+build.xml.data.CRC32=b8f3a06b
```
orogvany: I think you'll find that these types of files should be included in git exclude.

src/main/javafx/semux/Preloader/nbproject/project.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
orogvany: above you're using old logos, please update to newest logos

cryptokat:
Hello @Vjeredy 

Thank you very much for the pull request.

It would be really nice if you could make following improvements:

- [ ] Remove the legacy code of Swing GUI
- [ ] Move JavaFX source code to the package of `org.semux.gui`
- [ ] Load the JavaFX interface from [SemuxGui.java](https://github.com/semuxproject/semux/blob/develop/src/main/java/org/semux/gui/SemuxGui.java)
- [ ] Add test cases like https://github.com/semuxproject/semux/tree/develop/src/test/java/org/semux/gui
- [ ] Format the source code using command `mvn formatter:format` and `mvn license:format`
- [ ] Add third-party dependencies to https://github.com/semuxproject/semux/blob/develop/pom.xml

Vjeredy:
Regarding legacy Swing, I'm not sure why would you want to remove old GUI code before everything implemented properly in JavaFX. Because there might be features, pieces of code that can still come handy in the process.
Now about main application loader class - that one already means interaction with other modules of the project. Here you want to know as much as possible about your data flows, all the details, how exactly different modules are connected and that I do not know well.
Structure of data on the side of GUI part (as I see it) defined in Session which includes Accounts, Transactions, Delegates. Fields (properties actually) of these classes are bound to GUI elements of each scene. Can't tell for now if it compatible with what you have.
This is as far as I can go within this current pull request.

cryptokat:
@Vjeredy Thank you for your contribution. Ideally we would like to completely replace the Swing GUI with a JavaFX implementation. As you can see, there is a long list of reported GUI bugs. To make it more clear for developers on which part of code to look at, I believe we should keep only one version of GUI in this codebase. That's why I added removing Swing GUI in the todo list.

cryptokat:
Hi @semuxdev 
Should we merge the implementation into `javafx` branch then continue the development over there?

Vjeredy:
Somehow GUI tests cause maven build to fail, however tests pass when being run from IDE.

cryptokat:
Hello @Vjeredy
There is guide on how to run TestFX on Travis-CI: https://github.com/testfx/testfx/wiki/Travis-CI

cryptokat:
Hi @Vjeredy 
Would you mind if I merge this pull request into an independent branch `javafx` such that other developers can also help?

cryptokat:
Confirmed with @Vjeredy on Discord that we will continue the JavaFX development on branch `javafx`

cryptokat:
Manually pushed the changes to https://github.com/semuxproject/semux/tree/javafx

semuxdev:
Please rebase on dev branch and refine the commits according to the [Semux Contributing Guide](https://github.com/semuxproject/semux/blob/master/.github/contributing.md)

Further JavaFX pull request shall be merged to the `javafx` branch first.

cryptokat:
Squashed into one commit: https://github.com/semuxproject/semux/compare/develop...javafx

Issue #597, opened 2018-01-27
Address book doesn't open (related to #596)


kaar2:
This happens after doing one of the steps described in #596. After leaving an address with a blank name in the address book and closing the client. The address book won't open next time the client is launched.

#### System information

Semux version: v1.0.0

OS & Version: Windows 7 x64

#### Expected behaviour

After doing one of the steps described in #596, so the address book contains an address with a blank name and closing the client, the address book should open next time the client is launched.

#### Actual behaviour

The address book doesn't open next time the client is launched. It's fixed if you change the name of the blank named address to another name. However, this can only be done from the Receive tab, so if the address is removed from the wallet, there is no way to fix it without modifying the data files. The problem seems to be in the wallet.data file, as I tried to replace the whole client folder with a new one and the problem persisted. Only when I removed the wallet.data file the problem was fixed.

#### Steps to reproduce the behaviour

1. Add an address to the address book with a blank name.
2. Close the client.
3. Relaunch the client and click on the Address Book button.
4. Move the wallet.data from the client folder, restart the client and click on the button again.

My address is: 0xe9187ad84de23dc0e4eebc8dfb8ed1dc60072e22

semuxdev:
Same as #650

PR #650, opened 2018-02-09
GUI: Empty address book names

### LINKED ISSUE #596: Several bugs and unexpected behaviour of the address book
Yesterday I did some testing with the address book and found some unexpected behaviour. It happens if an address in the address book has an empty name. This can be caused by either one of the following:
1. Adding an address to the address book with an empty name.
2. Changing the name of an address in the address book to an empty name.
3. Adding a name to one of the addresses in the Receive tab and then removing the name.

Once one of the above happens the client becomes buggy. I'll the describe the issues I found, although I'm pretty sure there are more:
1. If you simply add an address to the address book with an empty name, the address won't actually show up in the address book, but the name will indeed change. In the drop lists in both Send and Delegates tabs the address will no longer appear and instead there will be an empty string (screenshots attached below).

2. If you change the name of an already existing address to the empty name, again the address book won't be updated (the address will appear there with the old name), but the new name will be updated everywhere else (screenshot attached below).

3. If you name an address through the Receive tab and then later want to remove the name (so the address will appear in the drop lists instead of the name) the only way to do that is by going to the address book and removing the address. When I tried to simply enter an empty name in the Receive tab, I got the same behaviour as in issue no. 2.

Additional comments:
- ~~The issues above happened not only when entering an empty string as a name. I also tried, space, tab, CR, LF and others. The null character however doesn't cause any issues.~~
I wasn't able to reproduce this with other strings beside the empty string. I checked the code and indeed it does occur only with the empty string.
- This is not related to the issues above , but when clicking on the Delete button in the address book there is no pop-up with "Are you sure?" text like there is when deleting an address in the Receive tab. If you delete your own address there is no problem, but after deleting an address that is not in your wallet data file, there is no way to restore it if you don't know the address.
#### System information

Semux version: v1.0.0

OS & Version: Windows 7, x64

My address is: 0xe9187ad84de23dc0e4eebc8dfb8ed1dc60072e22

orogvany:
There is inconsistent handling of address book names and empty
addresses.
This can cause old wallets to become broken, so empty names error
is relaxed in AddressBookEntry.  We still do our best to not allow
empty names.

Additionally, if a user selects cancel, we should not throw an 
error.

fixes #596, #597


orogvany:
Nevermind, looks like this was already done

PR #662, opened 2018-02-11
Config, Net: add mainnet.semux.net as an alternative dns seed

### LINKED ISSUE #624: Add more DNS lookup servers
DNS lookup for seed peers at startup has a few problems that should be addressed.

For a truly fault tolerant distributed system, there should not be single points of failure.

Currently, new clients cannot join the cluster if the DNS seed is down or attacked.  We should get some failover/secondary dns peer resolution servers up.  Likely this should be in the configuration, so we do not have to release new code in times of crisis, or as new ones are added.

I would imagine that semux.info, semuxchain.info, etc could host backup DNS resolution, preferrably pointing at a variety of known validators.

The code also assumes that the port will always be default, which is not the case.

cryptokat:
fix #624 

```
$ dig mainnet.semux.net @8.8.8.8

; <<>> DiG 9.9.5-9+deb8u12-Debian <<>> mainnet.semux.net @8.8.8.8
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 54467
;; flags: qr rd ra; QUERY: 1, ANSWER: 5, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
;; QUESTION SECTION:
;mainnet.semux.net.             IN      A

;; ANSWER SECTION:
mainnet.semux.net.      599     IN      A       23.102.239.10
mainnet.semux.net.      599     IN      A       35.198.39.94
mainnet.semux.net.      599     IN      A       35.201.187.136
mainnet.semux.net.      599     IN      A       35.226.132.132
mainnet.semux.net.      599     IN      A       51.143.156.145

;; Query time: 12 msec
;; SERVER: 8.8.8.8#53(8.8.8.8)
;; WHEN: Mon Feb 12 04:43:41 CST 2018
;; MSG SIZE  rcvd: 126

```

### DIFF
--- a/src/main/java/org/semux/net/NodeManager.java
+++ b/src/main/java/org/semux/net/NodeManager.java
@@ -21,6 +21,7 @@ import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Stream;
 
 import org.semux.Kernel;
 import org.semux.Network;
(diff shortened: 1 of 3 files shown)


src/main/java/org/semux/net/NodeManager.java
```diff
@@ -172,15 +173,17 @@ public int queueSize() {
...
-                    nodes.add(new Node(a, Constants.DEFAULT_P2P_PORT));
-                }
-            } catch (UnknownHostException e) {
-                logger.info("Failed to get seed nodes from {}", name);
-            }
-        }
+        names.parallelStream()
+                .map(name -> {
+                    try {
+                        return InetAddress.getAllByName(name);
```
orogvany: might advise adding a trim() to name. Or in the configuration parser, fixing up the current string.split call, as we don't want issues with folks adding more in future.

Issue #98, opened 2017-10-19
Validators with low forging ratio should be dropped


cryptozero:
Someone with enough voting power could introduce multiple stale validators and disrupt the network. Also an attack on several nodes, abandoned or shutdown validators could affect transaction processed by the network.

Validators with low performance should be droped from the top of validator list. After n blocks validators will be whitelisted again, if they fail again to perform, drop time doubles.

semuxproject:
The validator slashing mechanism needs more research and formal proof, as it introduces more state and may be exploited validators.

cryptozero:
One idea to improve health of network and reward best delegates:

- Less performing validator get dropped from list.
- Next delegate enter the list to forge.
- Dropped validator can enter the list again after one round is completed (100 blocks) in place of the less performing validator.
- If reentered validator is the dropped again in less than n blocks, drop time doubles.

This solution should incorporate some protection against DDoS targeting specific nodes.

AlexAltertrader:
I think you need a operation in the BAN in blockchain, issued by the validators working for the dead validator. After that, this validator is dropped from validators list for a ban period.

phash:
instant ban might not be the best solution... second chance before ban maybe?
It can always happen, a validator has just one bad luck block but is reliable all the other time. Look at the 90-99% validators. They are not bad validators, they just missed one block in a while..

So: second chance for x blocks - if the x blocks are over, the flag is removed. If the validator misses a second block, he gets a ban for y blocks and drops out of validators for that time span

AlexAltertrader:
@phash I agree, the second chance is necessary. And maybe the third. But no more, skipping the third round means that the validator has problems.

orogvany:
So in this case, what constituted a failed round? 100% failed forgings in the round?  Remember that voting is as important as forging, so I'd probably extend that to 'no votes/forgings in round', basically if the node is down.

Sadly, we do not currently keep track of voting information history in the blockchain I don't think.

AlexSemuxTrader:
I think that validators with a poor percentage of processed blocks should be discarded from the validators.
The maximum score window for processed blocks should be 300 blocks.
Then validators can be validators that have the next percent of the processed blocks.
For the first 10 blocks, -50%
For the first 100 blocks-80%
For the first 200 blocks-90%
For the first 300 blocks-95%

Issue #462, opened 2018-01-10
Validator % should be calculated only over last N blocks


orogvany:
#### Expected behaviour

Validator A - founded January
Validator B - founded April 1

Validator A goes down for 1 week in Feb.

It's April 2, validator B has only been up for 1 day.
Validator A has been up 100% since it's hitch in Feb when it forgot to upgrade software
Validator B now has a higher % than validator A, even though its only been up for 2 hours.  

I see this in mainnet.  If I see a validator with 80%, what do I do with that information if its a pool?
As a user, I'll see 100% and vote for it over someone that has been around since the start.

% validated over entire history is not relevant to current trust/uptime

A validator's % in UI should be recalculated over last N blocks to keep playing field even.

If not, I understand, but what are user's supposed to do with the validator %?  What's its purpose in UI?

semuxproject:
This would be a feature to implement after the first release.

orogvany:
I agree.  It's low pri

cryptokat:
related #98

PR #665, opened 2018-02-13
[WIP] DB: Store recent validator stats

orogvany:
Store the set of recent stats by block, so we can calculate
stats over a time range.

Currently this is defaulted to display last 7 days
in delegates panel.  But we can make this configurable

There's a lot of future work here how we present this
information.  Likely a new API method or parameter.

At 30 days, this should add ~2MB to the database
8 bytes * # blocks in 30 days * 3 stats possible

--
This is a work in progress to get feedback on methodology
in case there's a better way to get rolling stats, or some 
problem I'm not seeing.   Please pardon some cleanup necessary
with scrutinizer, etc while we figure out if this is otherwise agreeable  :D

Currently it does not backfill stats, but that can be added
once we figure out a methodology that works for folks.

So if you start with a current sync'd install, it will only show
stats from now going forward.  If you want to see it operate, 
can start a sync from scratch.

This has possible use for #98 as an easy entry point to get recent stats.

### DIFF
--- a/src/main/java/org/semux/core/Blockchain.java
+++ b/src/main/java/org/semux/core/Blockchain.java
@@ -9,6 +9,7 @@ package org.semux.core;
 import java.util.List;
 
 import org.semux.core.BlockchainImpl.ValidatorStats;
+import org.semux.core.BlockchainImpl.RecentValidatorStats;
 import org.semux.core.state.AccountState;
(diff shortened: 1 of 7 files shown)


cryptokat:
related: #462

semuxdev:
I will continue working on this, to make sure we have enough time testing the `v1.1.0`

PR #675, opened 2018-02-25
Consensus: Improve voting speed

orogvany:
Currently, each validator has a background thread
that takes in pending transactions and validates them.

When a block proposal comes in, the validator verifies each
transaction again.  As most validators will have most of the
pending transactions that the forger will have, optimize to
save double calculation.

This is noticed to improve block validation time by ~90%
On testnet validator for 5k transactions from around 2.2
seconds down to 20ms.

I'm not sure if with 100 validators if pending transactions
propagate fully, so the actual results may be less.

### DIFF
--- a/src/main/java/org/semux/consensus/SemuxBft.java
+++ b/src/main/java/org/semux/consensus/SemuxBft.java
@@ -14,6 +14,7 @@ import java.util.Set;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.locks.ReentrantReadWriteLock.WriteLock;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import org.semux.Kernel;
@@ -784,7 +785,7 @@ public class SemuxBft implements Consensus {
         }
 
         // [2] check transactions and results (skipped)
-        Set<Transaction> unvalidatedTransactions = getUnvalidatedTransactions(transactions);
+        List<Transaction> unvalidatedTransactions = getUnvalidatedTransactions(transactions);
 
         if (!Block.validateTransactions(header, unvalidatedTransactions, transactions, config.network())
                 || transactions.stream().mapToInt(Transaction::size).sum() > config.maxBlockTransactionsSize()) {
@@ -823,20 +824,19 @@ public class SemuxBft implements Consensus {
      * @param transactions
      * @return
      */
-    private Set<Transaction> getUnvalidatedTransactions(List<Transaction> transactions) {
+    private List<Transaction> getUnvalidatedTransactions(List<Transaction> transactions) {
 
-        // get all of the validated pending transactions
-        Set<Transaction> pendingUnvalidated = pendingMgr.getPendingTransactions(-1)
+        Set<Transaction> pendingValidatedTransactions = pendingMgr.getPendingTransactions(-1)
                 .stream()
                 .filter(pendingTx -> pendingTx.transactionResult.isSuccess())
                 .map(pendingTx -> pendingTx.transaction)
                 .collect(Collectors.toSet());
 
-        // filter out transactions we've already validated
-        Set<Transaction> unvalidatedTransactions = transactions
+        Predicate<Transaction> containsPredicate = pendingValidatedTransactions::contains;
+        List<Transaction> unvalidatedTransactions = transactions
                 .stream()
-                .filter(pendingUnvalidated::contains)
-                .collect(Collectors.toSet());
+                .filter(containsPredicate.negate())
+                .collect(Collectors.toList());
 
         logger.debug("Block validation: # txs = {}, # txs unvalidated = {} ms", transactions.size(),
                 unvalidatedTransactions.size());
--- a/src/main/java/org/semux/consensus/SemuxBft.java
+++ b/src/main/java/org/semux/consensus/SemuxBft.java
@@ -827,18 +827,14 @@ public class SemuxBft implements Consensus {
     private Set<Transaction> getUnvalidatedTransactions(List<Transaction> transactions) {
         Set<Transaction> unvalidatedTransactions = new HashSet<>(transactions);
 
-        int previouslyValidatedCount = 0;
-        List<PendingManager.PendingTransaction> pending = pendingMgr.getPendingTransactions(-1);
-        for (PendingManager.PendingTransaction t : pending) {
-            if (t.transactionResult.isSuccess()) {
-                boolean found = unvalidatedTransactions.remove(t.transaction);
-                if (found) {
-                    previouslyValidatedCount++;
-                }
+        List<PendingManager.PendingTransaction> pendingTransactions = pendingMgr.getPendingTransactions(-1);
+        for (PendingManager.PendingTransaction pendingTransaction : pendingTransactions) {
+            if (pendingTransaction.transactionResult.isSuccess()) {
+                unvalidatedTransactions.remove(pendingTransaction.transaction);
             }
         }
-        logger.debug("Block validation: # txs = {}, previously validated = {} ms", transactions.size(),
-                previouslyValidatedCount);
+        logger.debug("Block validation: # txs = {}, # txs unvalidated = {} ms", transactions.size(),
+                unvalidatedTransactions.size());
 
         return unvalidatedTransactions;
     }
--- a/src/main/java/org/semux/consensus/SemuxBft.java
+++ b/src/main/java/org/semux/consensus/SemuxBft.java
@@ -785,21 +785,7 @@ public class SemuxBft implements Consensus {
         }
 
         // [2] check transactions and results (skipped)
-
-        // remove any that are pending transactions we know about and have already
-        // validated
-        Set<Transaction> unvalidatedTransactions = new HashSet<>(transactions);
-
-        int previouslyValidatedCount = 0;
-        List<PendingManager.PendingTransaction> pending = pendingMgr.getPendingTransactions(-1);
-        for (PendingManager.PendingTransaction t : pending) {
-            if (t.transactionResult.isSuccess()) {
-                boolean found = unvalidatedTransactions.remove(t.transaction);
-                if (found) {
-                    previouslyValidatedCount++;
-                }
-            }
-        }
+        Set<Transaction> unvalidatedTransactions = getUnvalidatedTransactions(transactions);
 
         if (!Block.validateTransactions(header, unvalidatedTransactions, transactions, config.network())
                 || transactions.stream().mapToInt(Transaction::size).sum() > config.maxBlockTransactionsSize()) {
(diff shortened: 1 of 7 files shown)


src/main/java/org/semux/consensus/SemuxBft.java
```diff
@@ -812,6 +817,28 @@ protected boolean validateBlock(BlockHeader header, List<Transaction> transactio
...
 
+    /**
+     * Filter transactions to find ones that have not already been validated via the
+     * pending manager.
+     *
+     * @param transactions
+     * @return
+     */
+    private Set<Transaction> getUnvalidatedTransactions(List<Transaction> transactions) {
+        Set<Transaction> unvalidatedTransactions = new HashSet<>(transactions);
```
witoldsz: Just my subjective opinion, but I think it would be easier to follow what's happening if the order of actions was reversed, like this:

```java
Set<PendingManager.PendingTransaction> pendingUnvalidated = pendingMgr.getPendingTransactions(-1)
        .stream()
        .filter(t -> !t.transactionResult.isSuccess())
        .collect(toSet());

Set<Transaction> unvalidatedTransactions = transactions
        .stream()
        .filter(pendingUnvalidated::contains)
        .collect(toSet());
```

Now there is no need to deal with mutable state of collection, nested within `if` inside `for` loop.

src/main/java/org/semux/core/Transaction.java
```diff
@@ -289,4 +291,30 @@ public String toString() {
...
+        return new EqualsBuilder()
+                .append(encoded, that.encoded)
+                .append(hash, that.hash)
+                .append(signature, that.signature)
+                .isEquals();
+    }
+
+    @Override
+    public int hashCode() {
+        return new HashCodeBuilder(17, 37)
```
witoldsz: Since Java 1.7 it can be a one-liner: `return Objects.hash(encoded, hash, signature);`

witoldsz: ~Since Java 1.7 it can be a one-liner: `return Objects.hash(encoded, hash, signature);`~

Oops, my bad. The `Objects.hash` does not work with arrays, it will produce unique hash per array instance. Java is really trying hard to mislead developers :-/

orogvany: Yea. I had that at first but tests failed :/

src/main/java/org/semux/consensus/SemuxBft.java
orogvany: Sounds good. Will update

witoldsz: Also, when looking at the code, the `unvalidatedTransactions` need not to be a set anymore, simple list would do.

orogvany: I used a set as the lookup time is faster when querying it.

orogvany: I used a set as the lookup time is faster when querying it. O(1) vs O(n) afaik

witoldsz: Right, the lookup time (in large sets) is much faster when using hash tables. I was referring to the change that now the `unvalidatedTransactions` is not used to search for transactions anymore, but `pendingUnvalidated` is instead.

orogvany: 👍

src/main/java/org/semux/consensus/SemuxBft.java
```diff
@@ -812,6 +816,34 @@ protected boolean validateBlock(BlockHeader header, List<Transaction> transactio
...
+    /**
+     * Filter transactions to find ones that have not already been validated via the
+     * pending manager.
+     *
+     * @param transactions
+     * @return
+     */
+    private Set<Transaction> getUnvalidatedTransactions(List<Transaction> transactions) {
+
+        // get all of the validated pending transactions
```
witoldsz: this should be `// get all of the unvalidated pending transactions`
I propose to remove this comment, as it should be pretty obvious from the variable name what is this.

src/main/java/org/semux/consensus/SemuxBft.java
```diff
@@ -812,6 +816,34 @@ protected boolean validateBlock(BlockHeader header, List<Transaction> transactio
...
+        Set<Transaction> pendingUnvalidated = pendingMgr.getPendingTransactions(-1)
+                .stream()
+                .filter(pendingTx -> pendingTx.transactionResult.isSuccess())
```
witoldsz: looks like the `!` is missing here:
`.filter(pendingTx -> ! pendingTx.transactionResult.isSuccess())`

Otherwise this will be pending validated, not unvalidated.

src/main/java/org/semux/consensus/SemuxBft.java
```diff
@@ -812,6 +816,34 @@ protected boolean validateBlock(BlockHeader header, List<Transaction> transactio
...
+                .map(pendingTx -> pendingTx.transaction)
+                .collect(Collectors.toSet());
+
+        // filter out transactions we've already validated
```
witoldsz: This comment says what is almost literally written below.

src/main/java/org/semux/consensus/SemuxBft.java
```diff
@@ -812,6 +816,34 @@ protected boolean validateBlock(BlockHeader header, List<Transaction> transactio
...
+        Set<Transaction> unvalidatedTransactions = transactions
```
witoldsz: There is no need to create a `Set` here, simple `List` would do. Much less overhead when creating, and the only use of this is inside `Block.validateTransactions` which accepts a `Collection`.
`List<Transaction> unvalidatedTransactions = … .collect(toList());`

src/main/java/org/semux/consensus/SemuxBft.java
orogvany: Ah, I see where the confusion is.  Yes this checkin (and your original) are incorrect.

The point is to remove any pending transactions that have been validated from the list of the proposed block transactions.

Not to find intersection of pending that are unvalidated and also in pending.

The block proposer may have transactions that are not in our pending list, and that is fine.  But we can't ignore them.  I'll fix the logic here.

src/main/java/org/semux/consensus/SemuxBft.java
orogvany: 👍

src/main/java/org/semux/consensus/SemuxBft.java
witoldsz: OK, so the comment was right, but the variable name was wrong? 
Like this?
`pendingValidated = […].filter(it -> it.transactionResult.isSuccess())…`
and now we have to remove (`filter` with `!`) all the above from the initial `transactions` provided as an input:
`notValidated = transactions.[…].filter(it -> ! pendingValidated.contains(it))…`

orogvany: correct.

witoldsz: That would look even simpler if (wonderful) [jOOλ](https://github.com/jOOQ/jOOL) library would be used:
`notValidated = seq(transactions).removeAll(pendingValidated).toList();`

orogvany: Possibly, but introducing a new library feels heavy for this.

We could add it, but we'd want to audit the rest of our streams cases and migrate all of them I would think, which feels out of scope for this change.

Consistency across project is important.

witoldsz: Yes, adding jOOλ just for this could be an overkill. The truth is, it would help with other streams and also could help in places when imperative style is used because streams + lack of easy function composition in Java are short on features.

I am not trying to persuade to use it now. Just wanted to mention it. Maybe one day, you will be so tired of me mentioning it (here and there), that… :D

src/main/java/org/semux/consensus/SemuxBft.java
```diff
@@ -832,10 +832,9 @@ protected boolean validateBlock(BlockHeader header, List<Transaction> transactio
                 .map(pendingTx -> pendingTx.transaction)
                 .collect(Collectors.toSet());
 
-        Predicate<Transaction> containsPredicate = pendingValidatedTransactions::contains;
         List<Transaction> unvalidatedTransactions = transactions
                 .stream()
-                .filter(containsPredicate.negate())
+                .filter(it -> !pendingValidatedTransactions.contains(it))
```
witoldsz: This is so funny, I was playing the same game with poor's man functional programing in Java and decided for the `it -> !pendingValidatedTransactions.contains(it)` style when writing my comment as well :D

You know, jOOλ has this `not(Predicate<T>)` function, but it's `Seq` interface over streams has both `retain…` and `remove…` method families. Not to mention window functions, but it does not matter here anyway…

src/main/java/org/semux/consensus/SemuxBft.java
orogvany: I would be in favor of it if someone wanted to submit a pull request to clean up our streams usages with it :D (hint hint)

src/main/java/org/semux/consensus/SemuxBft.java
```diff
@@ -812,6 +816,32 @@ protected boolean validateBlock(BlockHeader header, List<Transaction> transactio
...
     }
 
+    /**
+     * Filter transactions to find ones that have not already been validated via the
+     * pending manager.
+     *
+     * @param transactions
+     * @return
+     */
+    private List<Transaction> getUnvalidatedTransactions(List<Transaction> transactions) {
```
cryptokat: @orogvany Would it make more sense moving this method to `PendingManager`? At my first glance it appears a little strange to have this private method in `SemuxBFT` as the method doesn't access any private/protected property other than `pendingMgr`.

src/main/java/org/semux/consensus/SemuxBft.java
```diff
@@ -812,6 +816,32 @@ protected boolean validateBlock(BlockHeader header, List<Transaction> transactio
...
+                .filter(pendingTx -> pendingTx.transactionResult.isSuccess())
+                .map(pendingTx -> pendingTx.transaction)
+                .collect(Collectors.toSet());
+
+        List<Transaction> unvalidatedTransactions = transactions
+                .stream()
+                .filter(it -> !pendingValidatedTransactions.contains(it))
+                .collect(Collectors.toList());
+
+        logger.debug("Block validation: # txs = {}, # txs unvalidated = {} ms", transactions.size(),
```
cryptokat: ` ms` here should be removed

src/main/java/org/semux/consensus/SemuxBft.java
orogvany: I agree that moving it might make more sense, but it really doesn't have anything to do with pending manager specifically.

I might make this method protected to match the rest of the class though, as even other helper methods like isFromValidator() are set as protected.

I am happy to move this method where ever it seems appropriate.

src/main/java/org/semux/consensus/SemuxBft.java
orogvany: done, thanks!

src/test/java/org/semux/consensus/SemuxBftTest.java
```diff
@@ -121,4 +124,49 @@ public void testDuplicatedTransaction() {
         // the block should be rejected because of the duplicated tx
         assertFalse(semuxBFT.validateBlock(block2.getHeader(), block2.getTransactions()));
     }
+
+    @Test
+    public void testFilterPendingTransactions() {
+        Key to = new Key();
+        Key from1 = new Key();
+        long time = System.currentTimeMillis();
+        Transaction tx1 = new Transaction(
```
witoldsz: The following almost 20 lines of code looks like a copy/paste from another test. Does not look good to me…

witoldsz: Also, the amount of information provided in all this is overwhelming, the code of this test should focus on the case. I read all this, and I have no idea what is relevant and what is irrelevant for the purpose it this test.

cryptokat:
Tested on the testnet with a few thousand TXs.

Issue #676, opened 2018-02-26
sometime the number is rounded wrong


likaigithub:
Hi dev,

#### System information

Semux version: v1.0.0

OS & Version: Windows 10

#### Expected behaviour

I send 0.1225 sem to someone.the amount should be 0.123 when it has 3 decimal.

#### Actual behaviour

it's 0.122.

#### Steps to reproduce the behaviour

FYI

![ozd bx a_ xm6t oic0_z i](https://user-images.githubusercontent.com/9426498/36655665-0df36d06-1aff-11e8-8b50-12d5d40f9eda.png)

PR #681, opened 2018-03-02
Support Customized Base Unit

### LINKED ISSUE #608: Allow user to choose the base unit
User should be able to select the base unit. Examples are `SEM`, `mSEM`, `μSEM` etc.

Low priority though

cryptokat:
fix #608, #676

### DIFF
--- a/src/main/java/org/semux/core/Unit.java
+++ b/src/main/java/org/semux/core/Unit.java
@@ -6,10 +6,14 @@
  */
 package org.semux.core;
 
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
 public class Unit {
     public static final long NANO_SEM = 1L;
 
-    public static final long MICRO_SEM = 1000L;
+    public static final long MICRO_SEM = 1_000L;
 
     public static final long MILLI_SEM = 1_000_000L;
 
@@ -18,6 +22,30 @@ public class Unit {
     public static final long KILO_SEM = 1_000_000_000_000L;
 
     public static final long MEGA_SEM = 1_000_000_000_000_000L;
+
+    public static final Map<String, Long> SUPPORTED;
+
+    static {
+        Map<String, Long> aMap = new HashMap<>();
+        aMap.put("μSEM", MICRO_SEM);
+        aMap.put("mSEM", MILLI_SEM);
+        aMap.put("SEM", SEM);
+        SUPPORTED = Collections.unmodifiableMap(aMap);
+    }
+
+    /**
+     * Converts an unit string into its decimal value.
+     *
+     * @param unit
+     * @return
+     */
+    public static long valueOf(String unit) {
+        if (SUPPORTED.containsKey(unit)) {
+            return SUPPORTED.get(unit);
+        } else {
+            throw new IllegalArgumentException(String.format("%s is an unsupported unit.", unit));
+        }
+    }
 
     private Unit() {
     }
--- a/src/main/java/org/semux/core/Unit.java
+++ b/src/main/java/org/semux/core/Unit.java
@@ -23,14 +23,28 @@ public class Unit {
 
     public static final long MEGA_SEM = 1_000_000_000_000_000L;
 
+    /**
+     * Mapping of unit string => base unit value.
+     */
     public static final Map<String, Long> SUPPORTED;
 
+    /**
+     * Mapping of unit string => number of digits after decimal point.
+     */
+    public static final Map<String, Integer> SCALE;
+
     static {
-        Map<String, Long> aMap = new HashMap<>();
-        aMap.put("μSEM", MICRO_SEM);
-        aMap.put("mSEM", MILLI_SEM);
-        aMap.put("SEM", SEM);
-        SUPPORTED = Collections.unmodifiableMap(aMap);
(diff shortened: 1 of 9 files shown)


src/main/java/org/semux/core/Unit.java
```diff
@@ -6,10 +6,14 @@
  */
 package org.semux.core;
 
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
 public class Unit {
```
witoldsz: What do you think about making this class an `enum`? It looks like enum and is used in the code as enum.

src/main/java/org/semux/gui/SwingUtil.java
```diff
@@ -327,23 +331,62 @@ public static String formatNumber(Number number) {
...
      * 
      * @param nano
+     * @param unit
+     * @param fractionDigits
      * @return
      */
-    public static String formatValue(long nano) {
-        return formatValue(nano, true);
+    public static String formatValue(long nano, String unit, int fractionDigits, boolean withUnit) {
+        return formatNumber(nano / (double) Unit.valueOf(unit), fractionDigits) + (withUnit ? " " + unit : "");
```
witoldsz: I am strongly advising not to "touch" `double` at any possible point when dealing with money or any other values with strict precision requirements. Use `BigDecimal` instead.

Floats and doubles are always approximate(*). Most of the time the approximation is good enough, but sometimes it's not and it can cause bugs very hard to track.

Check this out:

```
long i = 261797469654772356L;
double d = (double) i;
System.out.println(i);
System.out.println((long) d);

261797469654772356
261797469654772352
```

And it's just a taste of (scary) thing to come.

---
(*) One can see more approximation, in action, below:
```
System.out.println(1.0031d - 0.001d);
1.0021000000000002
```

src/main/java/org/semux/core/Unit.java
cryptokat: I agree it can be a good idea to refactor this class into an Enum, but I don't plan to mess up these constants in this pull request since they are being used in a lot of places.

src/main/java/org/semux/gui/SwingUtil.java
```diff
@@ -354,10 +404,20 @@ public static String formatValue(long nano) {
      * @throws ParseException
      */
     public static long parseValue(String str) throws ParseException {
-        if (str.endsWith(" SEM")) {
-            str = str.substring(0, str.length() - 4);
+        // check if there is a specific unit in the value string
+        for (Map.Entry<String, Long> unitMapping : Unit.SUPPORTED.entrySet()) {
+            if (str.endsWith(" " + unitMapping.getKey())) {
+                str = str.replace(" " + unitMapping.getKey(), "");
+                return BigDecimal.valueOf(parseNumber(str).doubleValue())
```
witoldsz: `BigDecimal` created of `double` is as useless as `double` itself.

The problem root is located in:
- `public static Number parseNumber(String str)` method produces a `Double` values,
- `public static String formatNumber(Number number, int decimals)` converts into `double` inside.

This entire class needs to be ripped of any `double` or `Number` or it is all for nothing.
I can do it in a separate pull request if you don't mind?

src/main/java/org/semux/core/Unit.java
witoldsz: True. I can make it so, in a separate pull request, to keep changes more focused on a single target.

src/main/java/org/semux/gui/SwingUtil.java
```diff
@@ -354,10 +404,20 @@ public static String formatValue(long nano) {
...
+                        .multiply(BigDecimal.valueOf(unitMapping.getValue()))
+                        .longValue();
+            }
         }
-        return (long) (parseNumber(str).doubleValue() * Unit.SEM);
+
+        // if not, use the default unit
+        return BigDecimal.valueOf(parseNumber(str).doubleValue())
```
witoldsz: Following 3 lines of code are a copy/paste from few lines above.
Check this out:

```java

String numberStr = Unit.SUPPORTED.keySet().stream()
    // check if there is a specific unit in the value string
    .filter(symbol -> str.endsWith(" " + symbol))
    .map(symbol -> str.replace(" " + symbol, ""))
    .findAny()
    // if not, use the default unit
    .orElse(str);

return BigDecimal.valueOf(parseNumber(numberStr).doubleValue())
    .multiply(BigDecimal.valueOf(Unit.valueOf(unit)))
    .longValue();
```

Now, there is explicit separation of string preparation for parsing, and it happens only once.

witoldsz: Oh crap, I've rushed a little bit. The parsing needs the proper unit:

```java
Pair<String, Long> numberUnit = Unit.SUPPORTED.entrySet().stream()
    .filter(e -> str.endsWith(" " + e.getKey()))
    .map(e -> Pair.of(str.replace(" " + e.getKey(), ""), e.getValue()))
    .findAny()
    .orElse(Pair.of(str, Unit.valueOf(unit)));

return BigDecimal.valueOf(parseNumber(numberUnit.getKey()).doubleValue())
    .multiply(BigDecimal.valueOf(numberUnit.getValue()))
    .longValue();
```

PR #690, opened 2018-03-03
UI: Remove File|Exit

### LINKED ISSUE #487: Reorganize Menu for better usability
#### System information

Semux version: v1.0.0-rc.5

OS & Version: Windows 10 64 bit

#### Expected behaviour

More user friendly menu

#### Actual behaviour

File -> Exit doesn't make any sense. You can add Exit as a button next to Lock/Unlock button
Menu buttons are too close. Add more space between menu items
Help should have at least basic How To Page

1. Open File Menu
2. Open Help Menu

orogvany:
In looking at other applications, File|Exit does not seem to match up.
there are other ways to shut down the application

This will keep our options cleaner.

fixes  #487

### DIFF
--- a/src/main/java/org/semux/gui/Action.java
+++ b/src/main/java/org/semux/gui/Action.java
@@ -53,8 +53,6 @@ public enum Action {
 
     CHANGE_PASSWORD,
 
-    EXIT,
-
     ABOUT,
 
     HELP,
--- a/src/main/java/org/semux/gui/MenuBar.java
+++ b/src/main/java/org/semux/gui/MenuBar.java
@@ -33,7 +33,6 @@ import org.semux.gui.dialog.ChangePasswordDialog;
 import org.semux.gui.dialog.ExportPrivateKeyDialog;
 import org.semux.gui.dialog.InputDialog;
 import org.semux.message.GuiMessages;
-import org.semux.util.SystemUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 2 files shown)


witoldsz:
Firefox has `File/Quit (Ctrl+Q)`, Chrome has `⋮/Exit`, all 3 of my IDEs have it, even system Calculator has `Calculator/Quit (Ctrl+Q)` menu.

What is wrong with it?

orogvany:
Hm..  I guess I can't speak for windows.  
In OSX, it's never under file.

It's just weird, as it's not a file operation, so why put it under file?  I assume everyone just X's or hotkeys programs to close these days.  I figured it was a legacy thing like using a floppy disk icon for save.

I could go either way, it's a minor thing.  If folks prefer working harder to close a program, then so be it :D

witoldsz:
Yes, OSX has everything different than Linux/Windows systems. For example Linux/Win Netbeans has File/Exit, but OSX Netbeans has no Exit there…

orogvany:
Sure, but it's still not a file operation.  Maybe System|Exit or similar?

For example Steam uses Steam|Exit, and as you mentioned chrome doesn't have a file menu.

witoldsz:
Lack of "exit" menu entry can possibly cause some troubles, when you run some awkward window manager, or when no window manager has launched (some docker images, installing XWindow on something like Ubuntu Server or minimal Debian and missing some packages?) and you have this app running inside some VNC and there is no window decorations… and now you want to exit and there is nothing you can click, no keyboard shortcut to press. Not a typical use-case, isn't it? :) `killall java` would do, for sure, though :)

orogvany:
How about Semux|Exit  then instead of File | Exit?

That brings it inline with chrome, calculator, and steam

witoldsz:
> How about Semux|Exit then instead of File | Exit?

So, instead of removing `File | Exit`, you want to rename `File` to `Semux`? Seems OK to me.

Issue #701, opened 2018-03-06
1.0.1 ipfilter.json Parsing Error


iiqqrs:
#### System information

Semux version: v1.0.1

OS & Version: Windows 10

#### Expected behavior

>Install microsoft redistributable 2012
>Install java 8
> run semux.exe, create new account and semux begins connecting to peers and syncing.

#### Actual behaviour

After installing java 8 and the 2012 redistributable, running semux shows the new account creation screen. As soon as the wallet.data is created and in the same directory, the error I see every time I load up semux is:

<img>https://i.imgur.com/LeYlYOv.png</img>  

#### Steps to reproduce the behaviour

1. Download semux 1.0.1 on windows 10. 
2. Install dependencies
3. Run semux

#### Backtrace

````
2018-03-06T18:46:01-0500 ERROR    Wallet           Failed to decrypt the wallet data
2018-03-06T18:46:04-0500 INFO     Kernel           Semux/v1.0.1-27d00eb/Windows/amd64
2018-03-06T18:46:04-0500 INFO     Kernel           System booting up: network = mainnet, networkVersion = 0, coinbase = bcf549a6b732d89c55a7b550850cc6ade41c298a
2018-03-06T18:46:04-0500 INFO     Kernel           Computer: manufacturer = HP, model = HP ENVY Notebook
2018-03-06T18:46:04-0500 INFO     Kernel           OS: name = Microsoft Windows 10 (Home) build 16299
2018-03-06T18:46:04-0500 INFO     Kernel           CPU: processor = Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz, cores = 4 / 8
2018-03-06T18:46:04-0500 INFO     Kernel           Memory: total = 12210 MB, available = 8454 MB, swap total = 1856 MB, swap available = 1856 MB
2018-03-06T18:46:04-0500 INFO     Kernel           Disk: name = \\.\PHYSICALDRIVE0, size = 244191 MB
2018-03-06T18:46:05-0500 INFO     Kernel           Network: name = Teredo Tunneling Pseudo-Interface, ip = []
2018-03-06T18:46:05-0500 INFO     Kernel           Network: name = Realtek PCIe GBE Family Controller, ip = []
2018-03-06T18:46:05-0500 INFO     Kernel           Network: name = Microsoft Wi-Fi Direct Virtual Adapter, ip = []
[...]

PR #703, opened 2018-03-07
GUI,CLI: improve error reporting of ipfilter.json parsing

cryptokat:
see #701

### DIFF
--- a/src/main/java/org/semux/cli/SemuxCli.java
+++ b/src/main/java/org/semux/cli/SemuxCli.java
@@ -45,7 +45,7 @@ public class SemuxCli extends Launcher {
             cli.setupLogger(args);
             // start
             cli.start(args);
-        } catch (LauncherException | ConfigException exception) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/semux/cli/SemuxCli.java
```diff
@@ -45,7 +45,7 @@ public static void main(String[] args) {
             cli.setupLogger(args);
             // start
             cli.start(args);
-        } catch (LauncherException | ConfigException exception) {
+        } catch (LauncherException | ConfigException | org.semux.net.filter.exception.ParseException exception) {
```
orogvany: import here shouldn't be fully qualified.

cryptokat: There's a duplicate class name on line 50

orogvany: Ah, sorry for not noticing.
Should we consider disambiguating the classname?

PR #707, opened 2018-03-07
Util: Reorder public ip lookup methods

cryptokat:
Some ISPs/routers do not return correct results from opendns resolvers due to DNS hijacking or DNS request filtering.
http://checkip.amazonaws.com should be queried at first considering that it provides higher coverage.

### DIFF
--- a/src/main/java/org/semux/util/SystemUtil.java
+++ b/src/main/java/org/semux/util/SystemUtil.java
@@ -104,7 +104,28 @@ public class SystemUtil {
      * @return an IP address if available, otherwise local address
      */
     public static String getIp() {
-        // [1] fetch IP address from OpenDNS. This works for socks5 proxy users.
+        // [1] fetch IP address from Amazon AWS. This works for public Wi-Fi users.
+        try {
+            URL url = new URL("http://checkip.amazonaws.com/");
+            URLConnection con = url.openConnection();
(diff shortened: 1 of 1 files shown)


orogvany:
Oh. Guess it resolves but sends wrong ip?

cryptokat:
@orogvany An user reports that his ISP redirects all unknown domain names to an advertised search engine.

orogvany:
Shouldnt that result in falling back to aws?  Or does it match regex just with wrong data?  Fix looks fine. Just wondering why it didnt work as intended.

cryptokat:
@orogvany The DNS hijacking doesn't make A lookup fail, instead it returns its own A record.

orogvany:
Ah. Anything we can do to id that?  Or just assume aws will make it?  More just out of curiosity.  Presumably someone could hijack aws the same way?

cryptokat:
@orogvany That would be out of scope of this pull request. In current implementation it requires a correct public IP advertised in handshake message for each semux node to successfully join the network. We may like to remove this restriction in the future for that kind of users

orogvany:
Fair enough

semuxdev:
We got more failure reports before when using amazon aws.

Many crypto users are under proxy/vpn, and the return of the Amazon AWS is not always correct in this case.

cryptokat:
Okay. Probably we should consider a better solution that can make the P2P server get rid of 3rd-party services.

PR #715, opened 2018-03-10
Core: Prune block votes

### LINKED ISSUE #712: Prune block votes
No need to store all block votes unless for voting analysis. Add option to prune block votes for regular users.

semuxdev:
Resolves #712


orogvany:
I agree delegates should be required to keep this info.

semuxdev:
Good points. Will close this pull request and move the pending manager updates to another PR.

PR #724, opened 2018-03-12
Core: amount is no longer a "long" type

witoldsz:
Using a generic type of "long" to represent an amount is sometimes convenient, but can easily cause a serious bug. For example: almost everywhere the `long` type represents an amount in "nano" semux, but __sometimes__ it is just a semux value, multiplied by 1*10e9 elsewhere.

New `Amount` class, representing a semux amount is safer to use.


witoldsz:
The `GUI: use BigDecimal instead of double for correctness` commit is here just temporarily, it'll get get removed once its branch gets merged.

src/main/java/org/semux/consensus/SemuxBft.java
```diff
@@ -876,11 +879,10 @@ protected void applyBlock(Block block) {
...
         // [5] apply block reward and tx fees
-        long reward = config.getBlockReward(number);
-        for (Transaction tx : block.getTransactions()) {
-            reward += tx.getFee();
-        }
-        if (reward > 0) {
+        Amount txsReward = block.getTransactions().stream().map(Transaction::getFee).reduce(ZERO, Amount::sum);
+        Amount reward = sum(config.getBlockReward(number), txsReward);
+        
+        if (reward.gte0()) {
```
witoldsz: fixme: `reward.gt0()`

src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -427,11 +430,10 @@ protected boolean validateBlockVotes(Block block) {
...
         // [5] apply block reward and tx fees
-        long reward = config.getBlockReward(block.getNumber());
-        for (Transaction tx : block.getTransactions()) {
-            reward += tx.getFee();
-        }
-        if (reward > 0) {
+        Amount txsReward = block.getTransactions().stream().map(Transaction::getFee).reduce(ZERO, Amount::sum);
+        Amount reward = sum(config.getBlockReward(block.getNumber()), txsReward);
+        
+        if (reward.gte0()) {
```
witoldsz: fixme: `reward.gt0()`

src/main/java/org/semux/core/Amount.java
```diff
@@ -0,0 +1,96 @@
...
+ * Copyright (c) 2017-2018 The Semux Developers
+ *
+ * Distributed under the MIT software license, see the accompanying file
+ * LICENSE or https://opensource.org/licenses/mit-license.php
+ */
+package org.semux.core;
+
+/**
+ *
+ * @author witoldsz
```
witoldsz: fixme: remove this line

orogvany:
I only see the premine/util that ever uses non nano sem, which is never utilized during runtime

witoldsz:
@orogvany 
> I only see the premine/util that ever uses non nano sem, which is never utilized during runtime

You are right. I though I have spotted it more that once.

orogvany:
It's possible I missed a spot, but devs are fairly careful to not mix/match units.  I'm surprised the premine stuff passed muster or hasn't been fixed at this point.  Anyway, having a very deterministic unit scaling like this I think has merit.

BTW,
Please check out https://github.com/semuxproject/semux/blob/master/.github/contributing.md

There's a few things you're missing, like prefacing CR name with component, and ordering of imports in the https://github.com/semuxproject/semux/blob/master/misc/intellij/guide.md section.

This will help the process go smoother

witoldsz:
@orogvany 
> few things you're missing, like prefacing CR name with component

What would be the component in this case?

orogvany:
Core for changes to the core data structures and algorithms

cryptokat:
@witoldsz Thank you for making this pull request! Would you please rebase this pull request on `develop` branch? All code changes should be made to `develop` branch before they can be merged into `master` branch.

witoldsz:
Oh, crap. Rebasing on develop caused conflicts – easy, but now tests are failing :-/ 
Will revisit what has changed.

src/main/java/org/semux/gui/SwingUtil.java
```diff
@@ -349,49 +351,43 @@ public static String formatValue(long nano) {
...
      * @param nano
+     * @param unit
      * @return
      */
-    public static String formatValueFull(long nano) {
-        return formatValue(nano, unit, 9, true);
+    private static String formatAmount(Amount a, Unit unit, int fractionDigits, boolean withUnit) {
+        return formatNumber(unit.toDecimal(a, fractionDigits), fractionDigits)
+                + (withUnit ? (" " + unit.symbol) : "");
     }
```
cryptokat: JavaDocs of these three methods are not up-to-date
```
String formatAmountNoUnit(Amount a)
String formatAmountFull(Amount a)
String formatAmount(Amount a, Unit unit, int fractionDigits, boolean withUnit)
```

witoldsz: Do we need JavaDocs with no content? What is the purpose of:
```java
    /**
     * Formats a Semux value.
     *
     * @param a
     * @return
     */
    public static String formatAmount(Amount a) {…}
```

The name of the method and the type of the parameter says everything. The JavaDoc adds nothing but a "noise".

Personally, when I see so JavaDocs like this, I unconsciously start to ignore them all, and then I miss also those which are supposed to provide help.

witoldsz: Can we just get rid of the JavaDocs when there is nothing there? That would make code easier to browse.

witoldsz: Fixed.

cryptokat: I think it would be fine to strip empty JavaDocs. In fact there is a lot of errors when I try to generate JavaDoc of semux by running `mvn javadoc:javadoc`. We will need to fix that once we have decided to publish the reference implementation as a library.

PR #725, opened 2018-03-13
Concensus: Testnet permits larger blocks

### LINKED ISSUE #706: [improvement] Testnet to not limit blocks to 1MB
For performance tweaking, it would be valuable if testnet would not limit block size to 1mb (~5k transactions).

I feel like we're eventually going to want to be able to support more transactions per block for higher throughput, so being able to see bottlenecks on the testnet and vet improvements there seems valuable.

Note: This limitation should remain in place on mainnet, only advise updating testnet to allow for pushing boundaries.  This will break consensus on testnet if we don't get validators to update, but I'm hoping we can just arrange for it, rather than forcing a fork.  Thoughts?

orogvany:
For testing purposes and performance tuning
it is desirable to be able to test larger amounts
of transactions than would normally make sense
in order to identify bottlenecks for future growth 
and optimization.

Note: This will affect concensus during rollout, but as it is testnet
it seems fine if during rollout that larger blocks will
fail validation until we get 67%.  The backup validator will
propose the smaller block and not hang concensus.

If we decide to go with larger blocks in future, we'll
have to use @cryptokat 's method for signaling forks,
but this is just for test purposes, and not proposing larger
blocks overall at this time.

fixes #706 

Also trivial rename for clarity.

### DIFF
--- a/src/main/java/org/semux/config/TestnetConfig.java
+++ b/src/main/java/org/semux/config/TestnetConfig.java
@@ -15,6 +15,8 @@ public class TestnetConfig extends AbstractConfig {
 
     public TestnetConfig(String dataDir) {
         super(dataDir, Network.TESTNET, Constants.TESTNET_VERSION);
+        // testnet allows a much larger block size for performance tuning (100MB)
+        maxBlockTransactionsSize = 100 * 1024 * 1024;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


semuxdev:
I feel like this is a little bit aggressive :)

orogvany:
That works for me.

PR #726, opened 2018-03-14
GUI: Add getBlockByNumber to console

### LINKED ISSUE #699: Get block by number in console
Is it possible to have get block by number command in the GUI console?

@orogvany

orogvany:
Add support for console API that isn't in REST
API.  As the console only supports string parameters
we add a getBlockByNumber.

This allows for additional console methods in the
future as well.

fixes #699


orogvany:
I could add support for other parameter parsing besides string and overload the method, but I think we're going to want to add additional console commands in the future, and don't want to be tied to rest api for everything we might add

PR #730, opened 2018-03-14
API: add a configuration 'api.returnNumbersAsStrings'

### LINKED ISSUE #729: API: Possible 64-bit Integer Overflow in JavaScript Clients
#### System information

Semux version: `v1.0.1`

#### Rationale 

Jackson [by default](https://github.com/FasterXML/jackson-databind/issues/911) serializes `Long` columns in plain numbers without considering the fact that JavaScript which is a popular client-side language, may not be able to deserialize 64-bit numbers correctly under some circumstances.

#### Demo of the Issue

Node.js demo which requests for an account's info then prints both of raw API response and deserialized json object:
```
> require('request-promise').get('http://user:pass@127.0.0.1:5171/get_account\?address\=0x541365fe0818ea0d2d7ab7f7bc79f719f5f72227&pretty=true').then(res => { console.log(res, JSON.parse(res)); });
> 
Raw:
{
  "success" : true,
  "result" : {
    "address" : "0x541365fe0818ea0d2d7ab7f7bc79f719f5f72227",
    "available" : 9399888811586101,
    "locked" : 300000000000000,
    "nonce" : 5111,
    "transactionCount" : 5121
  }
} 
Deserialized:
{ success: true,
  result: 
   { address: '0x541365fe0818ea0d2d7ab7f7bc79f719f5f72227',
     available: 9399888811586100,
     locked: 300000000000000,
     nonce: 5111,
     transactionCount: 5121 } }
```

The column of `result.available` is wrongly deserialized as `9399888811586100`

#### Proposed Fix

There are 2 proposed ways to fix this issue:

1. We may like to introduce an alternative `String` column for each `Long` column in API response. For example, to add a `available_str` column in string as an alternative of `available` column, as [Twitter has done](https://developer.twitter.com/en/docs/basics/twitter-ids) to its `id` column. In this way we can avoid breaking compatibility with existing 3rd-party applications. 
2. To return all numbers in `String`, which can be a breaking change

#### Credits

Thanks to @witoldsz for reporting this issue

cryptokat:
To fix #729 

### Common Changes

- [x] A new configuration property `api.returnNumbersAsStrings=true|false` is added to define whether the API server should return numbers as strings by default.
- [x] A new global GET parameter `returnNumbersAsStrings=true|false` is added for allowing API consumers to decide whether to get numbers as strings in runtime.
- [ ] Introduce API versioning in place of the configuration

### Breaking Change on Fresh Installs

`api.returnNumbersAsStrings` is enabled by default in `config/semux.properties` which may break existing API consumers.

### Change on Legacy Installs

`api.returnNumbersAsStrings` is disabled by default when the configuration is undefined in `config/semux.properties`, while a warning message that informs users about issue #729 will be shown in `debug.log` and console.


orogvany:
Are strings as numbers still swagger compliant?  I think they are but worth a check

orogvany:
Er nums as strings

orogvany:
Lets just introduce schema parameter (we'll need it anyway).

This prevents users from running 2 tools. One that uses strings and one that uses numbers.  

Schema introduction now will serve us better going forward

cryptokat:
Changed to WIP for introducing API versioning.

config/semux.properties
```diff
@@ -63,6 +63,10 @@ api.listenPort = 5171
 api.username = YOUR_API_USERNAME
 api.password = YOUR_API_PASSWORD
 
+# Whether to return numbers as strings.
+# This can be useful for clients who cannot deserialize 64-bit integers correctly, for example JavaScript.
+api.returnNumbersAsStrings = true
```
orogvany: with schema, doesn't this go away?

pom.xml
```diff
@@ -66,6 +66,7 @@
             </plugin>
 
             <!-- build swagger document -->
+            <!--
```
orogvany: delete commented out block

pom.xml
```diff
@@ -118,6 +119,56 @@
...
                 </dependencies>
             </plugin>
+            -->
+
+            <plugin>
+                <groupId>io.swagger</groupId>
+                <artifactId>swagger-codegen-maven-plugin</artifactId>
+                <version>2.3.1</version>
+                <executions>
+                    <!-- v1.0.1 -->
```
orogvany: shouldn't it be 1.0.0 for compatible version?

pom.xml
```diff
@@ -118,6 +119,56 @@
...
+                            <generateApis>false</generateApis>
+                            <generateSupportingFiles>false</generateSupportingFiles>
+                            <addCompileSourceRoot>true</addCompileSourceRoot>
+
+                            <output>${project.basedir}</output>
+                            <modelPackage>org.semux.api.v1_0_1</modelPackage>
+                        </configuration>
+                    </execution>
+
+                    <!-- v1.0.2 -->
```
orogvany: for semver, changes at this level should probably be 1.1.0 rather than 1.0.2

cryptokat:
Closed in favour of versioned API

PR #732, opened 2018-03-15
Merge master into develop

cryptokat:
#718 was mistakenly merged directly into `master` branch.

Maybe we should consider changing the default branch from `master` to `develop` to avoid this kind of mistake. See: https://help.github.com/articles/setting-the-default-branch/


witoldsz:
I am thinking: what is the reason for having a master and develop instead of just master with release tags on it? One branch seems to be easier to maintain than two, I guess.

cryptokat:
One advantage of this model is to deploy security patches quickly. If there's only a master branch receiving all the pull requests, there can be some backward incompatible changes which will prevent us from releasing an emergent security patch.

PR #734, opened 2018-03-15
Replace Scrutinizer-CI With FindBugs Maven Plugin

cryptokat:
to fix #733

PR #739, opened 2018-03-20
Util: Improve UTF8 CharSet Expression and Revert Scanner Default Charset

cryptokat:
Based on the comments of #734 

- [x] STDIN Scanner is reverted to system default encoding considering there might be users using non-UTF8 passwords
- [x] ~~~findbugs-exclude.xml filter ruleset is added to exclude `DM_DEFAULT_ENCODING` in `ConsoleUtil`~~~
- [x] `Charset.defaultCharset()` is explicitly specified as the charset of STDIN input stream to fix `DM_DEFAULT_ENCODING`

### DIFF
--- a/findbugs-exclude.xml
+++ b/findbugs-exclude.xml
@@ -1,8 +1,2 @@
 <FindBugsFilter>
-    <!-- It's fine for console reads to rely on default encoding -->
-    <Match>
-        <Class name="org.semux.util.ConsoleUtil"/>
-        <Bug pattern="DM_DEFAULT_ENCODING"/>
-    </Match>
 </FindBugsFilter>
-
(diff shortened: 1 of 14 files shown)


findbugs-exclude.xml
```diff
@@ -0,0 +1,8 @@
+<FindBugsFilter>
+    <!-- It's fine for console reads to rely on default encoding -->
+    <Match>
+        <Class name="org.semux.util.ConsoleUtil"/>
+        <Bug pattern="DM_DEFAULT_ENCODING"/>
```
witoldsz: As an alternative, you can consider removing this bug pattern exception and specify the default encoding explicitly.

`Scanner s = new Scanner(System.in, Charset.defaultCharset().name());`
or
`Scanner s = new Scanner(new InputStreamReader(System.in, Charset.defaultCharset()));`

It's not a crime to use default charset, the problem is that is's default and implicit, that is why it is considered a potential bug.

PR #741, opened 2018-03-20
Consensus: adding memoization to Vote#validate to avoid repeated validations

kaar2:
Currently, an incoming vote is validated 3 times (first time in onMessage, second time in onVote, and finally it is validated again when it's added to the VoteSet). To avoid that, a memoization is applied to Vote#validate. Forcing validation is possible by calling the overloaded Vote#validate with revalidate parameter set to true.

### DIFF
--- a/src/main/java/org/semux/consensus/SemuxBft.java
+++ b/src/main/java/org/semux/consensus/SemuxBft.java
@@ -638,7 +638,7 @@ public class SemuxBft implements Consensus {
             Vote vote = m.getVote();
 
             if (vote.getHeight() == height) {
-                if (vote.validate(true)) {
+                if (vote.revalidate()) {
                     events.add(new Event(Event.Type.VOTE, vote));
                 } else {
                     logger.debug("Invalid vote from {}", channel.getRemotePeer().getPeerId());
--- a/src/main/java/org/semux/consensus/SemuxBft.java
+++ b/src/main/java/org/semux/consensus/SemuxBft.java
@@ -638,7 +638,7 @@ public class SemuxBft implements Consensus {
             Vote vote = m.getVote();
 
             if (vote.getHeight() == height) {
-                if (vote.validate()) {
+                if (vote.validate(true)) {
                     events.add(new Event(Event.Type.VOTE, vote));
                 } else {
                     logger.debug("Invalid vote from {}", channel.getRemotePeer().getPeerId());
--- a/src/main/java/org/semux/consensus/Vote.java
+++ b/src/main/java/org/semux/consensus/Vote.java
@@ -6,6 +6,8 @@
  */
 package org.semux.consensus;
 
+import java.util.Optional;
+
 import org.semux.crypto.Key;
 import org.semux.crypto.Key.Signature;
 import org.semux.util.Bytes;
@@ -25,7 +27,7 @@ public class Vote {
 
     private byte[] encoded;
     private Signature signature;
-    private Boolean validated;
+    private Optional<Boolean> validated;
 
     public Vote(VoteType type, boolean value, long height, int view, byte[] blockHash) {
         this.type = type;
@@ -72,34 +74,33 @@ public class Vote {
      */
     public Vote sign(Key key) {
         this.signature = key.sign(encoded);
+        this.validated = Optional.empty();
         return this;
     }
 
     /**
-     * validate the vote format and signature.
+     * validate the vote format and signature while ignoring any cached validation
+     * value.
      * 
      * @return
      */
-    public boolean validate(boolean revalidate) {
-        if (revalidate) {
-            return (validated = type != null
-                    && height > 0
-                    && view >= 0
-                    && blockHash != null && blockHash.length == 32
-                    && encoded != null
-                    && signature != null && Key.verify(encoded, signature));
-        }
-        return validated != null ? validated
-                : (validated = type != null
-                        && height > 0
-                        && view >= 0
-                        && blockHash != null && blockHash.length == 32
-                        && encoded != null
-                        && signature != null && Key.verify(encoded, signature));
+    public boolean revalidate() {
+        return (validated = Optional.of(type != null
+                && height > 0
+                && view >= 0
+                && blockHash != null && blockHash.length == 32
+                && encoded != null
+                && signature != null && Key.verify(encoded, signature))).get();
     }
 
+    /**
+     * validate the vote format and signature. if exists, a memoized validation
+     * value is returned. NOTE: to force revalidation use {@link Vote#revalidate()}.
+     * 
+     * @return
+     */
     public boolean validate() {
-        return validate(false);
+        return validated.orElse(revalidate());
     }
 
     public VoteType getType() {
--- a/src/main/java/org/semux/consensus/Vote.java
+++ b/src/main/java/org/semux/consensus/Vote.java
@@ -102,7 +102,7 @@ public class Vote {
      * @return
      */
     public boolean validate() {
-        return validated.orElse(revalidate());
+        return validated.orElseGet(this::revalidate);
     }
 
     public VoteType getType() {
--- a/src/main/java/org/semux/consensus/Vote.java
+++ b/src/main/java/org/semux/consensus/Vote.java
@@ -35,6 +35,7 @@ public class Vote {
         this.height = height;
         this.view = view;
         this.blockHash = blockHash;
+        this.validated = Optional.empty();
 
         SimpleEncoder enc = new SimpleEncoder();
         enc.writeByte(type.toByte());
@@ -54,6 +55,7 @@ public class Vote {
         this.height = dec.readLong();
         this.view = dec.readInt();
         this.blockHash = dec.readBytes();
+        this.validated = Optional.empty();
 
         this.signature = Signature.fromBytes(signature);
     }
--- a/src/main/java/org/semux/consensus/Vote.java
+++ b/src/main/java/org/semux/consensus/Vote.java
@@ -25,6 +25,7 @@ public class Vote {
 
     private byte[] encoded;
     private Signature signature;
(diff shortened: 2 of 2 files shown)


src/main/java/org/semux/consensus/Vote.java
```diff
@@ -75,17 +76,38 @@ public Vote sign(Key key) {
...
+        if (revalidate) {
+            return (validated = type != null
+                    && height > 0
+                    && view >= 0
+                    && blockHash != null && blockHash.length == 32
+                    && encoded != null
+                    && signature != null && Key.verify(encoded, signature));
+        }
+        return validated != null ? validated
+                : (validated = type != null
```
witoldsz: Are the 5 following lines a copy/paste of the ones above?

src/main/java/org/semux/consensus/Vote.java
```diff
@@ -25,6 +25,7 @@
 
     private byte[] encoded;
     private Signature signature;
+    private Boolean validated;
```
witoldsz: Maybe `Optional` could be used instead?

src/main/java/org/semux/consensus/SemuxBft.java
```diff
@@ -638,7 +638,7 @@ public void onMessage(Channel channel, Message msg) {
             Vote vote = m.getVote();
 
             if (vote.getHeight() == height) {
-                if (vote.validate()) {
+                if (vote.validate(true)) {
```
witoldsz: Without a context, it is hard to guess why is this `validate(true)` should be used here. It's just the code does not "speak" for itself here, for me at least…

kaar2:
Yes I agree @witoldsz, the code was a bit messy. I cleaned it, and added some comments, hopefully it's more clear now.

orogvany:
Hm.. if its called 3 times, im not sure that adding this cached check is best way forward, rather than just reducing the calls for all code paths.

I havn't looked too much into this functionality, are there reasons for the lower checks, or do the higher up checks always get called?

kaar2:
@orogvany Well two of the three places where the vote is being validated are unrelated and removing the validation from one of them is not a good idea (onMessage method in SemuxBft class and addVote in VoteSet class). The only place where the validation might be unnecessary is in onVote method, since it's already validated in onMessage. At first I did pretty much what you suggest and simply removed the vote.validate() call from onVote. But as @cryptokat commented, it might be a better idea to just cache the validation instead of removing it from onVote and assume that every vote that is passed there is already validated.

src/main/java/org/semux/consensus/Vote.java
```diff
@@ -71,21 +72,37 @@ public static Vote newReject(VoteType type, long height, int view) {
      */
     public Vote sign(Key key) {
         this.signature = key.sign(encoded);
+        this.validated = null;
```
cryptokat: It's kind of weird to me assigning `null` to a boolean property.

orogvany: I'd probably prefer marking this as transient as well.

It's fine as it is, just more to inform intent.

src/main/java/org/semux/consensus/Vote.java
```diff
@@ -71,21 +72,37 @@ public static Vote newReject(VoteType type, long height, int view) {
...
+     * @param revalidate
+     * @return
+     */
+    public boolean validate(boolean revalidate) {
+        if (revalidate || validated == null) {
+            validated = type != null
+                    && height > 0
+                    && view >= 0
+                    && blockHash != null && blockHash.length == 32
+                    && encoded != null
```
witoldsz: Can `encoded` ever be null if it is always assigned in constructor? It should be `final` and checked for null when the vote gets created. I think this applies to other fields but signature as well.

witoldsz:
> But as @cryptokat commented, it might be a better idea to just cache the validation instead of removing it from onVote and assume that every vote that is passed there is already validated.

Would it be good enough to just cache the validation result within `vote`, without creating a second method? It's still unclear to me why caller needs to decide which method to call if all we need is to prevent excessive validation of already validated vote.

kaar2:
@witoldsz I'm struggling to understand if you are concerned about the functionality of the method or just about the code clarity. I find the method to be pretty clear given the documentation provided but maybe it's just me, If you feel that changing the overloaded method to a new method with a different name will be more clear I'll definitely do that.

About the functionality I think it a matter of perspective. While in theory, everything will work just fine without the ability to force revalidation, I still felt it made sense to allow that. When caching the validation one needs to be extremely cautious and make sure that the cached value is always up to date. It is very easy to forget about that and at some later point add a method that modifies the vote without updating the cached validation value (that's why I was uncertain whether caching the validation is way to go in the first place). Given that, I think it's a good idea to add the ability to enforce revalidation of a vote.

witoldsz:
> When caching the validation one needs to be extremely cautious and make sure that the cached value is always up to date.

We have a `Vote` instance. It's immutable in principle, i.e. there are no mutating methods and there should never be there. Once the `vote` is validated there is nothing what can possible make the validation obsolete because the function is 100% deterministic and there are no effects involved.

> It is very easy to forget about that and at some later point add a method that modifies the vote without updating the cached validation value

Objects are always designed in two ways: they are either meant to be mutable or they are immutable. E.g. `BigDecimal` is immutable. There is no question about it, no one would ever think about possibility to add mutations into it. `Vote` looks the same to me. It is a value object and making it mutable can have unforeseen consequences.

Having said that, I can see no reason why the `vote` is giving the choice to either validate or to re-validate. It makes the reasoning complicated and forces us to have a bigger picture of what is going on without a reason. Or I am missing something and just don't understand the bigger picture… This is why I was asking :)

kaar2:
If Vote was immutable there obviously wouldn't be any issue. Unfortunately it is mutable - signing the vote is allowed at any time. Moreover, you can create a new Vote instance without a signature. Now, if one validates a vote with an empty signature the validation value will be false. Without the possibility to revalidate it later, the validation value will always be false, even after the vote is signed. 

While this issue can specifically be dealt with by invoking the validate method when a vote is signed, I think the current solution is better for two reasons: First, it's unwise performance wise - the validation is an heavy task, which is why this fix was created in the first place. A vote needs to be validated only when necessary and not whenever it's signed. The second reason is the one I mentioned in my previous comment. Since it's not an immutable object by principle, other modifications might be implemented in a later stage which could cause the same problem.

cryptokat:
Hi @kaar2 
You can rewrite previous commit messages using interactive rebase. Check: https://help.github.com/articles/changing-a-commit-message/

src/main/java/org/semux/consensus/Vote.java
```diff
@@ -25,6 +27,7 @@
 
     private byte[] encoded;
     private Signature signature;
+    private Optional<Boolean> validated;
```
cryptokat: `validated` is not initialized, which causes NullPointerException

kaar2: Yes I just noticed that too. It's fixed now.

src/main/java/org/semux/consensus/Vote.java
```diff
@@ -71,21 +76,33 @@ public static Vote newReject(VoteType type, long height, int view) {
...
+    }
+
+    /**
+     * validate the vote format and signature. if exists, a memoized validation
+     * value is returned. NOTE: to force revalidation use {@link Vote#revalidate()}.
+     * 
+     * @return
+     */
+    public boolean validate() {
+        return validated.orElse(revalidate());
```
cryptokat: This expression should be changed to:
```
return validated.orElseGet(this::revalidate);
```
otherwise `Vote#revalidate` is called every time when `Vote#validate` is called, which makes the cache meaningless.

witoldsz:
> If Vote was immutable there obviously wouldn't be any issue. Unfortunately it is mutable - signing the vote is allowed at any time.

You are right, this is the part I was missing.

> Without the possibility to revalidate it later, the validation value will always be false, even after the vote is signed.

This is correct. This is what worries me the most. Let me explain…

> When caching the validation one needs to be extremely cautious and make sure that the cached value is always up to date.

Now I understand that we cannot simply cache the result of `validate()` blindly, because of what you have said above, but now I my concern is the very quote above. **It is that now – the caller has the responsibility to pick wisely: to validate or to revalidate.** If I choose to invoke `validate()` and that method was called earlier and then later someone has signed, the result will be wrong. Should I trace it back to figure out which method to call? What if someone changes something elsewhere? Will my part brake?

I think it is easier to challenge that problem within the `Vote` class, i.e. once we sign or change the signature, the cached result of validate has to be erased. That's it. If you encapsulate this simple logic within `Vote` everything is safe and under control. 

Your proposition makes the responsibility on caller, so if someone changes something what happens before the use of `validate()` – that change will make the system break elsewhere. It will be a bug much harder to track and much harder to guard.

What do you think?

kaar2:
>  It is that now – the caller has the responsibility to pick wisely: to validate or to revalidate. If I choose to invoke validate() and that method was called earlier and then later someone has signed, the result will be wrong. Should I trace it back to figure out which method to call? What if someone changes something elsewhere? Will my part brake?

I think we have a misunderstanding. There is no intention that validate and revalidate will produce different results. When a vote is signed, the cached value is indeed erased. However, to make sure this will always be the case, future changes, if implemented, must not break that, and this is not something I can't contol. For this reason, I added the option to force revalidation, for cases where you want to be extra cautious, as it reduces the chance for future bugs.

witoldsz:
> I added the option to force revalidation, for cases where you want to be extra cautious, as it reduces the chance for future bugs.

So, to sum it up: if someone asks:

**Q:** *when to use `validate()` and when `revalidate()`*

**A:** *As of now, they both work just fine, but in the future, when someone introduces a bug in `Vote` class, then you better be using `revalidate()`, because `validate()` will be broken.*

Is that right?

PR #742, opened 2018-03-21
Consensus: fix sync votes validation issue

semuxdev:


src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -406,26 +407,26 @@ protected boolean validateBlock(Block block, AccountState asSnapshot, DelegateSt
...
 
-            if (!set.contains(a) || !Key.verify(encoded, sig)) {
-                logger.debug("Invalid BFT vote: signer = {}", a);
-                return false;
-            }
+        // check validity of votes
+        if (!block.getVotes().stream()
+                .allMatch(sig -> Key.verify(encoded, sig) && validators.contains(Hex.encode(sig.getAddress())))) {
+            logger.debug("Block votes are invalid");
+            return false;
```
orogvany: Hm.. what would be a case where an invalid vote would make it into the block?

It feels like there's a case here, where an invalid vote will negate even having 2/3 proper voted consensus?

From a very naieve look at the logic anyway.

orogvany: This does feel like higher than a debug level event.

semuxdev: One could pass the validation by repeating one signature.

PR #755, opened 2018-03-29
Util: standardize system exit code

semuxdev:


src/main/java/org/semux/util/SystemUtil.java
```diff
@@ -47,6 +47,37 @@
...
+        public static final int FAILED_TO_LAUNCH_KERNEL = 204;
+
+        // database
+        public static final int FAILED_TO_OPEN_DB = 301;
+        public static final int FAILED_TO_REPAIR_DB = 302;
+        public static final int FAILED_TO_WRITE_BATCH_TO_DB = 303;
+
+        // upgrade
+        public static final int HARDWARE_UPGRADE_NEEDED = 401;
+        public static final int CLIENT_UPGRADE_NEEDED = 402;
```
cryptokat: Exit code must be between 0 and 255 under Linux platform, otherwise there can be an overflow. Ref: https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html

semuxdev: Thanks!

PR #764, opened 2018-04-05
Add v1.1 API and Swagger UI

cryptokat:
This pull request is a follow-up of #730.

### Backward Compatibility

This pull request introduces no backward-incompatibilies.

### For API Developers

v1.0.1 API is deprecated and expected to be removed in a future release, while it's still accessible at:
  - http://localhost:5171/get_info
  - http://localhost:5171/v1.0.1/get_info
  
v1.1.0 API is added and accessible at:
  - http://localhost:5171/v1.1.0/get_info

A bundled Swagger UI for v1.1 API is accessible at:
  - http://localhost:5171/v1.1.0/index.html

Java users can now initialize an API client directly by the SemuxApi interface using [JAXRSClientFactory](https://github.com/cryptokat/semux/blob/versioned-api/src/test/java/org/semux/api/v1_1_0/SemuxApiTestBase.java#L63). This removes the necessity of libraries like [semux-java-client](https://github.com/orogvany/semux-java-client).

### For Code Reviewers

The changes are mostly made to following folders:

- [src/main/java/org/semux/api](https://github.com/cryptokat/semux/tree/versioned-api/src/main/java/org/semux/api): API implementations
- [src/main/resources/org/semux/api](https://github.com/cryptokat/semux/tree/versioned-api/src/main/resources/org/semux/api): Swagger definition & UI

### Detailed Changes

- [x] Bump version number to `1.1.0` following [semver spec 7](https://semver.org/#spec-item-7)
- [x] Move current API implementation into its own package `org.semux.api.v1_0_1`
- [x] Deprecate v1.0.1 API
- [x] Create v1.1 swagger.json spec based on the v1.0.1 API spec with the following changes
  - [x] Split `/get_block` into 2 endpoints `/get_block_by_number?number` & `/get_block_by_hash?hash`
  - [x] Return all 64-bit integers as strings in order to address #729
  - [x] Implement restful http method mapping
    - `GET /add_node` => `POST /add_node`
    - `GET /add_to_blacklist` => `POST /add_to_blacklist`
    - `GET /add_to_whitelist` => `POST /add_to_whitelist`
    - `GET /create_account` => `POST /create_account`
    - `GET /delegate` => `POST /delegate`
    - `GET /send_transaction` => `POST /send_transaction`
    - `GET /sign_message` => `POST /sign_message`
    - `GET /transfer` => `POST /transfer`
    - `GET /unvote` => `POST /unvote`
    - `GET /vote` => `POST /vote`
- [x] Introduce Swagger Codegen to automatically generate API models & interfaces for v1.1 API
- [x] Refactor HttpHandler
- [x] Add bundled Swagger UI for API v1.1: http://localhost:5171/v1.1/index.html
- [x] Add Licenses

### Workflow Change on API Development

[swagger.json](https://github.com/cryptokat/semux/blob/versioned-api/src/main/resources/org/semux/api/v1_1_0/swagger.json) should now become the single source of truth of Semux API spec. API interfaces and models can be automatically generated from swagger.json with `mvn compile`.

This approach brings a few benefits:
- API interfaces and model classes are no longer needed to be maintained manually
- Swagger Doc/UI & 3rd party clients can rely on a static spec rather than a dynamically generated spec

### Future Development

- API v1.0.1 is deprecated in this pull request and should be removed in a future release.

- One obvious drawback of this this approach is that Swagger Codegen + JAX-RS introduces a lot of 3rd party dependencies. (almost 10 MB increasement of the jar size) 
In the future we may want to replace JAX-RS with an alternative which has minimal dependencies such as [Vert.x](https://mvnrepository.com/artifact/io.vertx/vertx-core/3.5.1) once v1.0.1 is removed.


witoldsz:
Operations like add_node, add_to_blacklist or any other side-effect actions are implemented with HTTP GET, but GET operations **must not** change data, it is only for retrieving.

cryptokat:
@witoldsz Thanks for the feedback. I just added one task for that.

src/gen/java/org/semux/api/v1_1_0/api/SemuxApi.java
```diff
@@ -0,0 +1,488 @@
...
+@Api(value = "/", description = "")
+public interface SemuxApi  {
+
+    /**
+     * Add node
+     *
+     * Adds a node to node manager.
+     *
+     */
+    @GET
```
witoldsz: `@POST`, never use `HTTP GET` for effects.

semuxdev:
- Could we use v1.1.0 instead of v1.0 for consistency?
- Since both v1.1.0 and v1.0.1 are supported, I'd suggest change the default to v1.0.1 (no need to add `/v1.1/` endpoint). The minimum change requirements for clients is to change the endpoint URL if they decide to upgrade kernel.
- The `src/gen` can be easily generated by `maven compile`. We don't need to include in the repo.
- Is it necessary to include all the dependencies? Only the generated Java/HTML code is required for distribution. The codegen code is only required for development (maven).

These are my initial thoughts on this PR. I will do code review later today.

cryptokat:
> Could we use v1.1.0 instead of v1.1 for consistency?

fixed

> Since both v1.1.0 and v1.0.1 are supported, I'd suggest change the default to v1.0.1 (no need to add /v1.1/ endpoint at the moment). The minimum change requirements for clients is to change the endpoint URL if they decide to upgrade kernel.

v1.0.1 is already the default API for unprefixed paths.

The routing works as:
- `/` => v1.0.1
- `/v1.0.1` => v1.0.1
- `/v1.1.0` => v.1.1.0

Ideally the API server should always require a version prefix to avoid confusion for API users. The unprefixed paths are planned to be removed along with v1.0.1 API as the sole purpose to keep them is to prevent the upgrade from breaking existing applications.


> The src/gen can be easily generated by maven compile as you mentioned. We don't need to include it in the repo.

added in .gitignore

> Is it necessary to include all the dependencies? Only the generated Java/HTML code is required for distribution. The codegen code is only required for development (maven).

Most of new dependencies are introduced by 2 mandatory packages `swagger-jaxrs` and `apache.cxf`. We will be able to get rid of them once we switch from JAXRS to an alternative like Vert.x
```
$ mvn dependency:tree -Dscope=compile

--- maven-dependency-plugin:2.8:tree (default-cli) @ semux ---
org.semux:semux:jar:1.1.0
+- javax.ws.rs:javax.ws.rs-api:jar:2.1:compile
+- io.swagger:swagger-annotations:jar:1.5.18:compile
+- io.swagger:swagger-jaxrs:jar:1.5.18:compile
|  +- io.swagger:swagger-core:jar:1.5.18:compile
|  |  +- com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:jar:2.9.3:compile
|  |  |  \- org.yaml:snakeyaml:jar:1.18:compile
|  |  +- io.swagger:swagger-models:jar:1.5.18:compile
|  |  \- javax.validation:validation-api:jar:1.1.0.Final:compile
|  +- javax.ws.rs:jsr311-api:jar:1.1.1:compile
|  +- org.reflections:reflections:jar:0.9.11:compile
|  |  \- org.javassist:javassist:jar:3.21.0-GA:compile
|  \- com.google.guava:guava:jar:20.0:compile
+- org.apache.cxf:cxf-rt-rs-client:jar:3.2.4:compile
|  +- org.apache.cxf:cxf-rt-transports-http:jar:3.2.4:compile
|  +- org.apache.cxf:cxf-core:jar:3.2.4:compile
|  |  +- com.fasterxml.woodstox:woodstox-core:jar:5.0.3:compile
|  |  |  \- org.codehaus.woodstox:stax2-api:jar:3.1.4:compile
|  |  \- org.apache.ws.xmlschema:xmlschema-core:jar:2.2.3:compile
|  \- org.apache.cxf:cxf-rt-frontend-jaxrs:jar:3.2.4:compile
+- javax.activation:activation:jar:1.1.1:compile
+- javax.xml.bind:jaxb-api:jar:2.3.0:compile
+- javax.annotation:javax.annotation-api:jar:1.3.1:compile
+- org.slf4j:slf4j-api:jar:1.7.25:compile
+- org.apache.logging.log4j:log4j-api:jar:2.10.0:compile
+- org.apache.logging.log4j:log4j-core:jar:2.10.0:compile
+- org.apache.logging.log4j:log4j-slf4j-impl:jar:2.10.0:compile
+- org.bouncycastle:bcprov-jdk15on:jar:1.59:compile
+- org.ethereum:leveldbjni-all:jar:1.18.3:compile
+- io.netty:netty-all:jar:4.1.22.Final:compile
+- org.xerial.snappy:snappy-java:jar:1.1.7.1:compile
+- com.github.oshi:oshi-core:jar:3.4.4:compile
|  +- net.java.dev.jna:jna-platform:jar:4.5.0:compile
|  |  \- net.java.dev.jna:jna:jar:4.5.0:compile
|  \- org.threeten:threetenbp:jar:1.3.6:compile
+- org.apache.commons:commons-lang3:jar:3.7:compile
+- org.apache.commons:commons-collections4:jar:4.1:compile
+- commons-validator:commons-validator:jar:1.6:compile
|  +- commons-beanutils:commons-beanutils:jar:1.9.2:compile
|  \- commons-digester:commons-digester:jar:1.8.1:compile
+- commons-logging:commons-logging:jar:1.2:compile
+- commons-cli:commons-cli:jar:1.4:compile
+- com.fasterxml.jackson.core:jackson-core:jar:2.9.5:compile
+- com.fasterxml.jackson.core:jackson-databind:jar:2.9.5:compile
|  \- com.fasterxml.jackson.core:jackson-annotations:jar:2.9.0:compile
+- com.fasterxml.jackson.jaxrs:jackson-jaxrs-json-provider:jar:2.9.5:compile
|  +- com.fasterxml.jackson.jaxrs:jackson-jaxrs-base:jar:2.9.5:compile
|  \- com.fasterxml.jackson.module:jackson-module-jaxb-annotations:jar:2.9.5:compile
+- net.i2p.crypto:eddsa:jar:0.2.0:compile
+- com.google.zxing:core:jar:3.3.2:compile
+- org.bitlet:weupnp:jar:0.1.4:compile
+- com.github.zafarkhaja:java-semver:jar:0.9.0:compile
\- com.github.ben-manes.caffeine:caffeine:jar:2.6.2:compile
------------------------------------------------------------------------
```

> Many of the generated imports are either unused or unnecessary. Also, the java.xml.bind is deprecated in Java 9 and will be removed in future release.

Following deprecated dependencies are added exclusively for Java 9 runtime. They are mandatory for Java 9 JVM as they are no longer included by default:
- javax.activation:activation
- javax.xml.bind:jaxb-api
- javax.annotation-api:javax.annotation-api

We either need to maintain our own swagger codegen templates or require those dependencies. Maintaining our own templates isn't something trivial and it will probably prevent us from upgrading to later versions of swagger codegen.

semuxdev:
Sorry, I meant default to the new version v1.1.0

semuxdev:
It works as described.

orogvany:
@witoldsz On the GETs modifying data -> I don't know that I agree that it *must* never break that.  For a CRUD type RESTful interface, sure, but there's countless examples of successful APIs to the contrary.  The HTTP spec defines it as best a 'convention' and uses 'should'.  the HTTP2 spec removes the sections about 'safe' methods entirely IIRC.

LICENSE_3RD
```diff
@@ -19,20 +32,33 @@ EdDSA-Java:
...
-Jackson:
+FasterXML Jackson:
     https://github.com/FasterXML/jackson
     Apache License, Version 2.0
 
+FasterXML Stax2 API:
+    https://github.com/FasterXML/stax2-api
+    BSD License
+
+Google Guava:
```
orogvany: Previously when I had a commit (#461) it was noted that guava was unacceptable for this project due to 
"Lightweight (library like guava/spring should be avoided because it's too heavy; they also depend on a lot of other libraries)"

semuxdev: Hi @orogvany, I'm expecting a lot of these dependencies will be removed as we deprecate JAXRS

cryptokat: It's implicitly required as a dependency of swagger-jaxrs

witoldsz:
The GET /add_node is not about REST-whatever. The GET methods are not supposed to modify data, so the "infrastructure" like PROXY or libraries are (in general, with exceptions) free to retry or cache as they wish. Also, when a developer looks at the API and see GET like this, it does not feel right.

Thanks for accepting the change to POST. The "What-on-Earth??" effect is no more when looking at the spec :smile:

src/main/java/org/semux/api/v1_0_1/SemuxApi.java
```diff
@@ -13,28 +13,28 @@
...
-import org.semux.api.response.GetTransactionLimitsResponse;
-import org.semux.api.response.GetTransactionResponse;
-import org.semux.api.response.GetValidatorsResponse;
-import org.semux.api.response.GetVoteResponse;
-import org.semux.api.response.GetVotesResponse;
-import org.semux.api.response.ListAccountsResponse;
-import org.semux.api.response.SendTransactionResponse;
-import org.semux.api.response.SignMessageResponse;
-import org.semux.api.response.VerifyMessageResponse;
+import org.semux.api.v1_0_1.response.AddNodeResponse;
```
witoldsz: We previously had no versioning, now we have new, incompatible, and the version is still `1.something`?
If we consider the old API was like `0.x` then why is this not a `1.0`?
If we consider the old API was like `1.0` then why is this not a `2.0`?

witoldsz: Oooops, I did not realize this IS the old API, it's just the package change :)

src/main/java/org/semux/api/Version.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package org.semux.api;
+
+import org.semux.util.exception.UnreachableException;
+
+public enum Version {
+
+    v1_0_1,
+
+    v1_1_0;
```
witoldsz: Should we have a `2.x`? This new API is not compatible with the old one (numbers wrapped in quotes, HTTP GET to POST changes).

cryptokat: This is going to be a backward-compatible release. According to [semver](https://semver.org/#spec-item-7), only `Y` should be increased.

src/main/resources/org/semux/api/v1_1_0/swagger.json
```diff
@@ -0,0 +1,1773 @@
...
+                    }
+                },
+                "security" : [
+                    {
+                        "basicAuth" : [ ]
+                    }
+                ]
+            }
+        },
+        "/get_account" : {
```
witoldsz: Can we remove all the `/get_…` prefixes from GET requests?
`GET /get_account/` vs `GET /account`
`GET /block_by_hash` vs `GET /block_by_hash`, etc…

src/main/java/org/semux/api/Version.java
witoldsz: I though the change is backward compatible, because now we have versioning, so new API 2.x is new, but everyone can still use the `1.0.1` which is as it was?

cryptokat: Yes, that makes more sense to have API's own versioning numbers considering that we are not going to change the API version for every single release of semux wallet. I'll make a pull request for that.

PR #767, opened 2018-04-06
DB: Store recent validator stats

### LINKED ISSUE #98: Validators with low forging ratio should be dropped

semuxdev:
This is primarily based on #665, to resolve #98 and #462.


cryptokat:
Profiling result:
selection_346

cryptokat:
Do we really need to store `RecentValidatorStats` in the database? Considering that most of overhead comes from encoding & decoding of `RecentValidatorStats` data structure, we can actually store the stats in a memory cache as `RECENT_VALIDATOR_STATS_DURATION` is kept low.

src/main/java/org/semux/core/BlockchainImpl.java
```diff
@@ -493,10 +505,24 @@ public ValidatorStats getValidatorStats(byte[] address) {
         return (value == null) ? new ValidatorStats(0, 0, 0) : ValidatorStats.fromBytes(value);
     }
 
+    @Override
+    public RecentValidatorStats getRecentValidatorStats(byte[] address) {
+        ByteArray key = ByteArray.of(address);
+
+        RecentValidatorStats stats = recentValidatorStats.getIfPresent(key);
+        if (stats == null) {
+            stats = new RecentValidatorStats();
```
cryptokat: @semuxdev 
Sorry for the wrong information. My previously suggested approach doesn't really work well as there isn't an efficient way to recover stats at a certain height. 

I just come up with an alternative approach that could work. Please let me draft another pull request for that.

semuxdev: Agreed, it's also my concern.

I was thinking of adding the following index
```
block_number => primary validator
```,
so we can compute the recent stats easily.

semuxdev:
Deprecated until we have better design.

Issue #766, opened 2018-04-06
API add create/sign raw transaction


semuxdev:
I was wondering if we need to add following APIs:

- Add `create_raw_transaction`
- Add `sign_raw_transaction`
- Rename `send_transaction` to `broadcast_raw_transaction`

These are very usefull for offline signing

cryptokat:
resolved by #771

PR #771, opened 2018-04-08
Several API Improvements 

cryptokat:
### Rename API 1.1.0 to API 2.0.0

Considering that we are not going to bump API version for every single release of Semux wallet, it makes more sense to let API have its own versioning numbers following semver.

As the v1.1.0 API introduces a few incompatible changes with v1.0.1 API, it should be named as v2.0.0 instead.

### Improve Naming of API 2.0 Endpoints

Http request method & naming of API 2.0 endpoints are improved following REST semantic web standard.

### Add /compose-raw-transaction & /sign-raw-transaction

to address #766


orogvany:
This should make it easier to implement the securities for POSTs on #579

src/main/java/org/semux/core/Transaction.java
```diff
@@ -238,6 +238,38 @@ public long getTimestamp() {
...
+        return new Transaction(
+                Network.of(networkId),
+                TransactionType.of(type),
+                to,
+                value,
+                fee,
+                nonce,
+                timestamp,
+                data);
+    }
```
yulongaion: Can we abstract this decoding logic into an method, shared by the constructor and this method?

semuxdev: Can we abstract this decoding logic into a method, shared by the constructor and this method?

cryptokat: fixed by https://github.com/semuxproject/semux/pull/771/commits/c0244233b2eb643a975007ecf302cbf191155c4b

PR #778, opened 2018-04-09
Report Database Upgrading Progress on Splash Screen

### LINKED ISSUE #757: New splash screen should account for DB upgrade
with the new splash screen and 1.0.2 requiring a longrunning db upgrade
it currently shows full progress bar while doing upgrade
it would be nice if this showed progress so users do not think app is hung

cryptokat:
to fix #757 

- [x] add an internal pubsub service
- [x] report database upgrading progress on splash screen


src/main/java/org/semux/event/PubSub.java
```diff
@@ -0,0 +1,169 @@
...
+    /**
+     * Start the {@link this#eventProcessingThread}.
+     *
+     * @throws UnreachableException
+     *             this method should only be called for once, otherwise an
+     *             exception will be thrown.
+     */
+    private synchronized void start() {
+        if (!semaphore.tryAcquire()) {
+            throw new UnreachableException("PubSub service can be started for only once");
```
semuxdev: Would `AtomicBoolean` fit better here?

src/main/java/org/semux/event/PubSub.java
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        eventProcessingThread.start();
```
semuxdev: Can we name the event processing thread?

src/main/java/org/semux/event/PubSub.java
```diff
@@ -0,0 +1,169 @@
...
+                    event = queue.take();
+                } catch (InterruptedException e) {
+                    interrupt();
+                    return;
+                }
+
+                ConcurrentLinkedQueue<PubSubSubscriber> q = subscribers.get(event.getClass());
+                if (q != null) {
+                    for (PubSubSubscriber subscriber : q) {
+                        subscriber.onPubSubEvent(event);
```
semuxdev: Please consider wrapping in safe execute (or catch exceptions).

src/main/java/org/semux/event/PubSub.java
```diff
@@ -0,0 +1,169 @@
...
+
+    private PubSub() {
+        queue = new LinkedBlockingQueue<>();
+        subscribers = new ConcurrentHashMap<>();
+        eventProcessingThread = new EventProcessingThread();
+        semaphore = new Semaphore(1);
+    }
+
+    public static PubSub getInstance() {
+        return instance;
```
semuxdev: I don't like the singleton-pattern design here. `PubSub` event handing can be generic and reused elsewhere.

cryptokat:
@semuxdev Raised issues have been addressed in the latest commits.

PR #781, opened 2018-04-11
API: include COINBASE transaction in block response

cryptokat:
This makes it easier for API consumers to account for block rewards.

### Tasks

- [x] BlockchainImpl: add `TYPE_COINBASE_TRANSACTION_HASH` index
- [x] API/TypeFactory: include COINBASE tx in block type


semuxdev:
Not sure if we should expose this transaction to public. If you recall, the signer of the coinbase transaction is from a magic account.

PR #782, opened 2018-04-11
Wallet: Update security of wallet with bcrypt

### LINKED ISSUE #758: Security weakness in wallet encryption
The user-supplied password is used in a single round of Blake2b to generate the aes key used for encrypting the private key in the wallet.dat file:
`byte[] key = Hash.h256(Bytes.of(password));`
As Blake2b is even faster than SHA1 or MD5, it means that millions of passwords a second can be tried on a given wallet.dat file; this means that any password likely chosen by a normal user would be easily brute-forced if their wallet.dat file was stolen.

Compare that to the Bitcoin wallet.dat encryption:
It also uses AES for the actual encryption, but uses at least 25000 rounds of hashing to get the key:
```
                if (pMasterKey.second.nDeriveIterations < 25000)
                    pMasterKey.second.nDeriveIterations = 25000;
```

This means that the Semux wallet.dat is significantly (several orders of a magnitude) less resistant to  brute force attacks on the password.

Suggested fix:
Increase the number of rounds of hashing to at least 100 000, as hardware is faster, and do not allow users to enter passwords shorter than 10 characters. Ideally, the passwords should also be checked against known breach lists such as https://haveibeenpwned.com/Passwords

orogvany:
Increment wallet version to add bcrypt hashing to generate
password for aes key.  This takes about 330ms on fairly
standard hardware.

fixes #758

### DIFF
--- a/src/main/java/org/semux/core/Wallet.java
+++ b/src/main/java/org/semux/core/Wallet.java
@@ -19,6 +19,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 
+import org.bouncycastle.crypto.generators.BCrypt;
 import org.semux.core.exception.WalletLockedException;
 import org.semux.crypto.Aes;
 import org.semux.crypto.CryptoException;
@@ -37,7 +38,8 @@ public class Wallet {
     private static final Logger logger = LoggerFactory.getLogger(Wallet.class);
 
     private static final int VERSION = 3;
-    private static final int HASH_ROUNDS = 100_000;
+    private static final int SALT_LENGTH = 16;
+    private static final int BCRYPT_COST = 12;
 
     private File file;
     private String password;
@@ -112,10 +114,8 @@ public class Wallet {
                     newAliases = readAddressAliases(key, dec);
                     break;
                 case 3:
-                    // to discourage brute forcing, use many rounds of hashing
-                    for (int i = 0; i < HASH_ROUNDS; i++) {
-                        key = Hash.h256(key);
-                    }
+                    byte[] salt = dec.readBytes();
+                    key = BCrypt.generate(Bytes.of(password), salt, BCRYPT_COST);
                     newAccounts = readAccounts(key, dec, true, version);
                     newAliases = readAddressAliases(key, dec);
                     break;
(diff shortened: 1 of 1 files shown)


semuxdev:
I was also wondering if the public key in the wallet can be removed.

orogvany:
We could remove the public key.  Is the idea there that the attacker wouldn't be able to check which keys were present?

semuxdev:
Yep. The intention was to separate the keys so you don't need to unlock all accounts. It seems to be less necessary, since the whole wallet uses the same password.

src/main/java/org/semux/core/Wallet.java
```diff
@@ -104,10 +105,18 @@ public boolean unlock(String password) {
...
                     break;
                 case 2:
-                    newAccounts = readAccounts(key, dec, true);
+                    newAccounts = readAccounts(key, dec, true, version);
+                    newAliases = readAddressAliases(key, dec);
+                    break;
+                case 3:
+                    // to discourage brute forcing, use many rounds of hashing
+                    for (int i = 0; i < HASH_ROUNDS; i++) {
+                        key = Hash.h256(key);
```
cryptokat: Would it be more secure to use [PBKDF2](https://stackoverflow.com/a/27928435/670662) with a random salt?

cryptokat: Also there is [BCrypt](https://github.com/bcgit/bc-java/blob/master/core/src/test/java/org/bouncycastle/crypto/test/BCryptTest.java#L3:43) considered as GPU-resistant.

orogvany: I'm good with either bcrypt or PBKDF2, let me know which one you folks prefer, and I'll move ahead with it.

cryptokat: I am saying to replace the loop of h256 with PBKDF2 or BCrypt. PBKDF2 and BCrypt are purposely designed for password encryption utilizing multi-iteration hashing. We should reuse those algorithms instead of implementing our own. 

The salt should be a cryptographically secure random number that is updated & stored for every encryption.

orogvany: Yup, got it. ;)  Thought I caught my edit in time.  So which do you prefer?

cryptokat: :+1:

src/main/java/org/semux/core/Wallet.java
```diff
@@ -104,10 +107,16 @@ public boolean unlock(String password) {
...
+                    newAccounts = readAccounts(key, dec, false, version);
                     break;
                 case 2:
-                    newAccounts = readAccounts(key, dec, true);
+                    newAccounts = readAccounts(key, dec, true, version);
+                    newAliases = readAddressAliases(key, dec);
+                    break;
+                case 3:
+                    byte[] salt = dec.readBytes();
+                    key = BCrypt.generate(Bytes.of(password), salt, BCRYPT_COST);
```
cryptokat: h256 at Line 100 is redundant for version 3

PR #783, opened 2018-04-11
 GUI: Notify user when there are updates 

### LINKED ISSUE #754: checkVersionLoop should inform when a new version is available
It might be nice to check latestVersion from api.semux.org
And inform users when a new version is available for download.

We currently just error if the minVersion is out of date.

orogvany:
Update the loop for checking versions to
1) check at first startup, then every 8 hours, instead
of once every 5 minutes
2) Add check against latest to inform user when a new
version has been posted, but only notify them once per
session.

fixes #754

### DIFF
--- a/src/main/java/org/semux/gui/SemuxGui.java
+++ b/src/main/java/org/semux/gui/SemuxGui.java
@@ -76,6 +76,7 @@ public class SemuxGui extends Launcher {
     private JFrame main;
     private Thread dataThread;
     private Thread versionThread;
+    private String lastVersionNotified;
 
     public static void main(String[] args) {
         try {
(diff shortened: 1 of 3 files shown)


src/main/java/org/semux/gui/SemuxGui.java
```diff
@@ -336,14 +337,19 @@ public synchronized void stop() {
...
             try {
-                Thread.sleep(5L * 60L * 1000L);
-
-                // compare version
-                String v = getMinVersion();
-                if (v != null && SystemUtil.compareVersion(Constants.CLIENT_VERSION, v) < 0) {
+                Version v = getCurrentVersions();
+                if (v != null && SystemUtil.compareVersion(Constants.CLIENT_VERSION, v.minVersion) < 0) {
                     JOptionPane.showMessageDialog(null, GuiMessages.get("WalletNeedToBeUpgraded"));
                     SystemUtil.exitAsync(SystemUtil.Code.CLIENT_UPGRADE_NEEDED);
```
cryptokat: Should it `return;` here?

src/main/java/org/semux/gui/SemuxGui.java
```diff
@@ -336,14 +337,19 @@ public synchronized void stop() {
     protected void checkVersionLoop() {
         while (!Thread.currentThread().isInterrupted()) {
             try {
-                Thread.sleep(5L * 60L * 1000L);
```
cryptokat: Should we keep the timeout of 5 minutes for `minVersion` check? The previous implementation appears to intentionally set this short timeout in order to inform outdated users asap.

orogvany: The original check was a little weird.  It would let you run with an outdated/broken version, but once you ran it for 5 minutes, then it would tell you, then shut down.  It seemed better to put the sleep after the first check.

Since we notify on startup that there is a new version available, they should get that notification well before it stops working, or if they havn't run it awhile, they'll get notification on startup.

I'm ok with the every 5 minute check, but as semux grows in size, every user making a call every 5 minutes seems a little excessive.  I figured a longer check seemed a little more reasonable, but I can set it back to 5 minutes if you'd wish.  I can also go back to waiting 5 minutes before notification, but I think I prefer early notification

semuxdev: The original design is due to lack of consideration. All API services are implemented with Java Sevlets; the current load is under capability.

PR #787, opened 2018-04-11
Consensus: add restrictions on coinbase & recipient

### LINKED ISSUE #786: Block coinbase can be any address
Would this break anything other than the validator stats?

cryptokat:
fix #786

### Added Restrictions

- [x] BFT & Sync: Disallow `Constants.COINBASE_KEY` 
- [x] BFT: Block#coinbase should always equal to `proposal.getSignature().getAddress()`
- [x] BFT & Sync & PendingManager: Disallow sending txs to coinbase magic account

### DIFF
--- a/src/main/java/org/semux/consensus/SemuxSync.java
+++ b/src/main/java/org/semux/consensus/SemuxSync.java
@@ -379,6 +379,7 @@ public class SemuxSync implements SyncManager {
         // Added checks to UNIFORM_DISTRIBUTION fork:
         // - blocks should never be forged by coinbase magic account
         // - transactions should never be sent to coinbase magic account
+        // TODO: move checks to Transaction#validate after fork activation
         if (chain.forkActivated(block.getNumber(), UNIFORM_DISTRIBUTION)) {
             if (Arrays.equals(block.getCoinbase(), Constants.COINBASE_KEY.toAddress())) {
                 logger.warn("A block forged by the coinbase magic account is not allowed");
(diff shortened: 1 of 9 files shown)


semuxdev:
To maintainers: please verify the whole chain to make sure no one has broken these rules; it needs to be effective as soon as possible once it's merged.

cryptokat:
@semuxdev done and tested on the mainnet & testnet

src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -373,6 +376,22 @@ protected boolean validateBlock(Block block, AccountState asSnapshot, DelegateSt
         BlockHeader header = block.getHeader();
         List<Transaction> transactions = block.getTransactions();
 
+        // Added checks to UNIFORM_DISTRIBUTION fork:
+        // - blocks should never be forged by coinbase magic account
+        // - transactions should never be sent to coinbase magic account
+        if (chain.forkActivated(block.getNumber(), UNIFORM_DISTRIBUTION)) {
```
cryptokat: Just a reminder these checks in sync manager will only be effective once `UNIFORM_DISTRIBUTION` is activated. Otherwise a single transaction will be able to freeze the whole network.

semuxdev: NOTE: after the fork is activated, let's move the transaction `from`/`to` validation to `Transaction#Validate()`

PR #796, opened 2018-04-13
Util: adjust test log levels

semuxdev:
The CI logs are too long to read. Time to clean up.

Still haven't figured out how to disable logs from `KernelRule`

### DIFF
--- a/src/test/resources/log4j2-test.xml
+++ b/src/test/resources/log4j2-test.xml
@@ -7,17 +7,17 @@
         </Console>
     </Appenders>
     <Loggers>
-        <Logger name="org.semux.api" level="DEBUG" />
-        <Logger name="org.semux.core" level="DEBUG" />
-        <Logger name="org.semux.consensus" level="DEBUG" />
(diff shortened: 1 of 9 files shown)


src/test/resources/log4j2-test.xml
```diff
@@ -7,17 +7,17 @@
...
-        <Logger name="org.semux.net" level="DEBUG" />
-        <Logger name="org.semux.vm" level="DEBUG" />
-        <Logger name="org.semux.wrapper" level="DEBUG" />
+        <Logger name="org.semux.api" level="INFO" />
+        <Logger name="org.semux.core" level="INFO" />
+        <Logger name="org.semux.consensus" level="INFO" />
+        <Logger name="org.semux.db" level="INFO" />
+        <Logger name="org.semux.net" level="INFO" />
+        <Logger name="org.semux.vm" level="INFO" />
+        <Logger name="org.semux.wrapper" level="INFO" />
```
cryptokat: This can make it very inconvenient during development from my experience. I think we should ignore this file if developers have different preferences on logging level of test case execution.

Issue #704, opened 2018-03-07
API significantly slower after 1.0.1 release


orogvany:
I've noticed that response time from the API has slowed down significantly since 1.0.1 release

I'm guessing this is due to getAccountTransactions now returning block number, but it could be another change.  I just use this call a lot.

I can continue to look into which call is slower, but it looks like some call is 10x slower or so

orogvany:
looks like this can be closed due to fixed issue.

PR #797, opened 2018-04-13
Improper initialization of HttpHandler causing serious API performance degrade

cryptokat:
`Map<Version, ApiHandler> apiHandlers` should be initialized for only once.

Related: #704

### DIFF
--- a/src/main/java/org/semux/api/http/HttpHandler.java
+++ b/src/main/java/org/semux/api/http/HttpHandler.java
@@ -21,12 +21,13 @@ import java.net.URI;
 import java.nio.charset.Charset;
 import java.security.MessageDigest;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import javax.activation.MimetypesFileTypeMap;
 import javax.ws.rs.core.Response;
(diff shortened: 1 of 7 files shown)


src/main/java/org/semux/api/http/HttpHandler.java
```diff
@@ -106,6 +104,7 @@ public HttpHandler(Kernel kernel) {
      */
     protected HttpHandler(Config config, ApiHandler apiHandler) {
         this.config = config;
+        this.apiHandlers = new ConcurrentHashMap<>();
```
witoldsz: Why `ConcurrentHashMap` if it is never modified?

src/main/java/org/semux/api/http/SemuxApiService.java
```diff
@@ -47,11 +50,17 @@ public Thread newThread(Runnable r) {
...
 
+    private final Map<Version, ApiHandler> apiHandlers;
+
     String ip;
     int port;
 
     public SemuxApiService(Kernel kernel) {
         this.kernel = kernel;
+
+        this.apiHandlers = new ConcurrentHashMap<>();
```
witoldsz: Same question here: why `ConcurrentHashMap`?

src/main/java/org/semux/api/http/HttpHandler.java
cryptokat: Indeed `ConcurrentHashMap` isn't necessary here. Fixed in https://github.com/semuxproject/semux/pull/797/commits/5dd158db0a558afe8c73eaa15ad21ca09dc3943d

semuxdev:
I don't think bundle version and handler is a good idea. There may exist multiple implementations for the same version (probably from test cases and future R&D)

cryptokat:
@semuxdev Good point. Fixed in https://github.com/semuxproject/semux/pull/797/commits/3a3276e63b48ed4b1709228325c8c5c14dc38630

PR #819, opened 2018-04-22
Add Java 10 compatibility 

semuxdev:
- [ ] Deprecate Java 9 support (Public Updates ended on March 2018)
- [ ] Add Java 10/11 tests


semuxdev:
Re-submit later

PR #827, opened 2018-05-01
Fix memory leak issue

### LINKED ISSUE #826: Memory leak
screenshot from 2018-04-30 21-34-30
screenshot from 2018-04-30 21-36-17

semuxdev:
Resolves #826 

- [x] Fix vote cached validation result
- [x] Fix connection counter
- [ ] ~~Fix log event piles up~~

### DIFF
--- a/src/main/java/org/semux/net/ConnectionLimitHandler.java
+++ b/src/main/java/org/semux/net/ConnectionLimitHandler.java
@@ -54,11 +54,10 @@ public class ConnectionLimitHandler extends ChannelInboundHandlerAdapter {
     @Override
     public void channelInactive(ChannelHandlerContext ctx) throws Exception {
         super.channelInactive(ctx);
-
-        AtomicInteger cnt = connectionCount
-                .getIfPresent(((InetSocketAddress) ctx.channel().remoteAddress()).getAddress());
-        if (cnt != null) {
-            cnt.decrementAndGet();
+        InetAddress address = ((InetSocketAddress) ctx.channel().remoteAddress()).getAddress();
+        AtomicInteger cnt = connectionCount.get(address, k -> new AtomicInteger(0));
+        if (cnt.decrementAndGet() < 0) {
+            connectionCount.invalidate(address);
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/semux/net/ConnectionLimitHandler.java
```diff
@@ -41,7 +44,7 @@ public ConnectionLimitHandler(int maxConnectionsPerIp) {
     public void channelActive(ChannelHandlerContext ctx) throws Exception {
         super.channelActive(ctx);
         InetAddress address = ((InetSocketAddress) ctx.channel().remoteAddress()).getAddress();
-        if (connectionCount.computeIfAbsent(address, k -> new AtomicInteger(0))
+        if (connectionCount.get(address, k -> new AtomicInteger(0))
```
witoldsz: When cache goes beyond `maximumSize`, the connectionCount for given address will drop down to zero.
Is that okay?

semuxdev: Given that the LRU cache is sufficiently large and the keys are `InetAddress`, this should be fine. However, we may need to invalid a counter if its value goes below zero.

semuxdev:
@witoldsz @cryptokat @orogvany any idea why the log events pile up?

cryptokat:
@semuxdev How long does it take to reproduce the memory leak of Log4j? `Log4jLogEvent` objects are normally garbage-collected as I observed in VisualVM.

semuxdev:
It’s very hard to reproduce. The dump we got was from one seed node after running months.

orogvany:
Hm, why would logger hold reference to VoteSet?  Are we passing Voteset objects into logger as param and trying to rely on tostring formatting or some such?  

Or am I misunderstanding the logger memory leak?

cryptokat:
@orogvany Check the heap dump in https://github.com/semuxproject/semux/issues/826 where ` Log4jLogEvent` objects occupies 31 megabytes of memory.

orogvany:
I don't see the heap dump posted, just some screen shots which doesnt include the expanded Log4JLogEvent?  Can someone post or send me the heapdump?

src/main/java/org/semux/net/ConnectionLimitHandler.java
witoldsz: The question is: why is the LRU cache used for `connectionCount`? LRU does not seem to fit well here.

PR #838, opened 2018-05-07
API: add 'GET /account/votes?address'

### LINKED ISSUE #836: API: no endpoint to fetch all my votes
The 
- `GET` `/v2.0.0/delegates` has `votes` as a number, but there is no data of who was voting.

Also, the 
- `GET` `/v2.0.0/account` has no info of votes. 

The only way to display the "Votes from Me" column in "Delegates" section, using API, is to query 
- `GET` `/v2.0.0/votes?delegate=X` for each delegate, one by one. That is 130 requests on mainnet.

cryptokat:
fix #836


src/main/java/org/semux/api/v2_0_0/TypeFactory.java
```diff
@@ -73,6 +76,18 @@ public static DelegateType delegateType(BlockchainImpl.ValidatorStats validatorS
...
 
+    public static List<AccountVoteType> accountVotes(Blockchain blockchain, byte[] address) {
+        DelegateState delegateState = blockchain.getDelegateState();
+        return delegateState
+                .getDelegates()
+                .parallelStream()
+                .map(delegate -> new AccountVoteType()
+                        .delegate(TypeFactory.delegateType(blockchain.getValidatorStats(delegate.getAddress()), delegate))
+                        .votes(String.valueOf(delegateState.getVote(address, delegate.getAddress()).getNano())))
+                .filter(accountVote -> !accountVote.getVotes().equals("0"))
```
witoldsz: Why do you first create an `AccountVoteType` and when it is created, you are getting rid of it when votes are `"0"`?

PR #854, opened 2018-05-13
Consensus: optimize SemuxSync#validateBlock

cryptokat:
fix #803 

A local benchmark with Intel E3-1231 + JVM 8 shows that about 60 seconds of CPU time is saved per 10K blocks with the optimization.


cryptokat:
@orogvany makes a good point. An attacker will be able to easily disrupt the network by broadcasting invalid blocks which are not signed by validators. Also there is no easy way to revert the local blockchain db before a Trie db is implemented.

PR #855, opened 2018-05-15
API: remove BlockType#date in favour of BlockType#timestamp

cryptokat:
~~~ISO-8601 is a broadly supported date format by multiple languages. For example, in Javascript a `Date` object can be directly constructed using an ISO-8601 string, while the original format of `YYYY-MM-DD` isn't supported.~~~


witoldsz:
I did not notice there are both: `timestamp` and `date` fields.
I would suggest removing the excessive `date` field: it is just the `timestamp` value in different format.

cryptokat:
I am agreed with @witoldsz `date` field is kind of redundant.

semuxdev:
I have been expecting this removal. :)

PR #857, opened 2018-05-16
GUI: Add Consensus Info Table

### LINKED ISSUE #852: GUI: Add Consensus Info Table at Home Panel
Quoting from https://github.com/semuxproject/semux/issues/816#issuecomment-387607779

- [x] Current primary validator
- [x] Next primary validator
- [x] Next validator set update
- ~~~Number of YES/NO votes~~~

cryptokat:
fix #852 

- [x] add consensus table on the home panel
- [x] add test case

### DIFF
--- a/src/test/java/org/semux/gui/panel/HomePanelTest.java
+++ b/src/test/java/org/semux/gui/panel/HomePanelTest.java
@@ -67,10 +67,10 @@ public class HomePanelTest extends AssertJSwingJUnitTestCase {
         // mock wallet model
         WalletDelegate primaryValidator = new WalletDelegate(
                 new Delegate(new Key().toAddress(), "primary".getBytes(), 0, Amount.ZERO));
-        when(walletModel.getPrimaryValidator()).thenReturn(Optional.of(primaryValidator));
+        when(walletModel.getPrimaryValidatorDelegate()).thenReturn(Optional.of(primaryValidator));
         WalletDelegate nextPrimaryValidator = new WalletDelegate(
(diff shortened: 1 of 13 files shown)


src/test/java/org/semux/gui/panel/HomePanelTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+        // mock kernel
+        KernelMock kernelMock = spy(kernelRule1.getKernel());
+        application = GuiActionRunner.execute(() -> new HomePanelTestApplication(walletModel, kernelMock));
+
+        window = new FrameFixture(robot(), application);
+        window.show().requireVisible().moveToFront();
+        window.label("primaryValidator").requireVisible().requireText("primary");
+        window.label("nextPrimaryValidator").requireVisible().requireText("next");
+        window.label("nextValidatorSetUpdate").requireVisible().requireText(GuiMessages.get(
```
orogvany: This should use same date format as "Block Time" (with seconds added)

otherwise, looks good.  Looks a little weird with each validator on own line, rather than continuing row/column layout from Overview

orogvany:
screen shot 2018-05-18 at 12 05 39 am

cryptokat:
@orogvany 
Updated layout according to suggestions:
selection_020

`Attempt Number` isn't something that is directly available to non-validator nodes as those nodes don't participate in the SemuxBFT consensus.

orogvany:
Thats fine.  I was just thinking of other fields that might be useful.  Would tell non validators that concensus was 'stuck' at the moment and explain why a transaction might not be going through.

I like the new layout!  looks good to me

Issue #830, opened 2018-05-04
API: fetch transactions for multiple accounts


witoldsz:
As of now, it is not possible to get a bulk of transactions for multiple accounts.
For example: display last 100 transactions of 3 accounts.

As of now, we have to fetch 3x100 transactions, merge and sort them and pick top/bottom 100.

Expected: API endpoint to fetch X first or X last transactions for all the accounts provided, e.g.:
`GET` `/account/last-transactions?size=100&address=add1,addr2,addr3`

witoldsz:
Turns out there is no way to implement pagination without reading, at least, `size+offest` or `size*page` records.

Also, to read transactions from multiple accounts in a chronological order, there must be an algorithm implemented like this:
1. read last tx from each account,
2. add them to a result list,
3. add them to special list ordered by timestamp
4. read next tx from account whose tx is number one in special list
5. add that tx to result list
6. replace that tx with its predecessor on special list, sort it
7. go to step 4 until there is either nothing more to read or the result is big enough
8. sort result list

There is also a simpler way: read `size*page` from all accounts, concat, sort and cut the piece.

Either way: what if someone takes 10 accounts with biggest history and asks for 100 records on page 4444… With the crazy algorithm above, it will cause `444400` tx reads, with the simple one it will be 10x more and all this to return just 100 transactions.

This is all crazy or it is too late for me to think clear.

witoldsz:
We need a special index for this or we can forget about reading and paginating transaction history of several accounts at once.

witoldsz:
Mu current solution to this problem looks like this:
![image](https://user-images.githubusercontent.com/544581/39835817-b14c580e-53d1-11e8-9cb1-63d1befde3b6.png)

That would be nice if we could just ask for history of few accounts merged and sorted, but as of now I can see no easy implementation.

Maybe it's not that bad to have history segregated by account?

orogvany:
I'm not sure how we'd go about this with the current levelDb

Take a look at BlockchainImpl.getTransactions()
The keys are        
` return Bytes.merge(Bytes.of(TYPE_ACCOUNT_TRANSACTION), address, Bytes.of(n));`
where n is the transaction number of the account (think nonce)
On the server side, currently we'd have to do as you are having to do, grab all transactions for all accounts, sort, and then paginate.

mdodong:
Can you do something like,

Array AccountsList
for account in AccountList
    Get account transactions 

sort
paginate

orogvany:
[...]

PR #860, opened 2018-05-18
GUI: Add transactions filtering

### LINKED ISSUE #550: Add filters to transactions panel
#### System information

Semux version: 1.0.0

#### Expected behaviour

As I run a validator, I have tons of invotes/coinbase transactions
There's 1000s of transactions shown.

I would like to be able to see transactions filtered by either
Type
From
To

#### Actual behaviour

1000s of transactions that are hard to find what I'm looking for

orogvany:
It's very difficult to locate a given transaction currently,
particularly for pool operators or heavy usage users.

Add filterable dropdowns to transactions panel.

fixes #550

### DIFF
--- a/src/main/java/org/semux/gui/panel/TransactionsPanel.java
+++ b/src/main/java/org/semux/gui/panel/TransactionsPanel.java
@@ -391,8 +391,7 @@ public class TransactionsPanel extends JPanel implements ActionListener {
         }
 
         // update filters that are not set for reduced filter set if filter not already
-        // set on them for
-        // further filtering
+        // set on them for further filtering
         if (to == null) {
             toModel.setData(allTo.stream()
                     .map(it -> new ComboBoxItem<>(SwingUtil.describeAddress(gui, it.getData()), it.getData()))
--- a/src/main/java/org/semux/gui/panel/TransactionsPanel.java
+++ b/src/main/java/org/semux/gui/panel/TransactionsPanel.java
@@ -288,8 +288,8 @@ public class TransactionsPanel extends JPanel implements ActionListener {
      */
     protected void refresh() {
         transactions.clear();
-        Set<byte[]> to = new HashSet<>();
-        Set<byte[]> from = new HashSet<>();
+        Set<ByteArray> to = new HashSet<>();
+        Set<ByteArray> from = new HashSet<>();
 
         // add pending transactions
         transactions.addAll(gui.getKernel().getPendingManager().getPendingTransactions()
@@ -328,8 +328,8 @@ public class TransactionsPanel extends JPanel implements ActionListener {
         // track all used addresses
         for (StatusTransaction statusTransaction : transactions) {
             Transaction transaction = statusTransaction.getTransaction();
-            to.add(transaction.getFrom());
-            from.add(transaction.getTo());
+            to.add(new ByteArray(transaction.getFrom()));
+            from.add(new ByteArray(transaction.getTo()));
         }
 
         // filter transactions
@@ -338,11 +338,11 @@ public class TransactionsPanel extends JPanel implements ActionListener {
         tableModel.setData(filteredTransactions);
 
         fromModel.setData(from.stream()
-                .map(it -> new ComboBoxItem<>(SwingUtil.describeAddress(gui, it), it))
+                .map(it -> new ComboBoxItem<>(SwingUtil.describeAddress(gui, it.getData()), it.getData()))
                 .collect(Collectors.toCollection(TreeSet::new)));
 
         toModel.setData(to.stream()
-                .map(it -> new ComboBoxItem<>(SwingUtil.describeAddress(gui, it), it))
+                .map(it -> new ComboBoxItem<>(SwingUtil.describeAddress(gui, it.getData()), it.getData()))
                 .collect(Collectors.toCollection(TreeSet::new)));
 
         if (tx != null) {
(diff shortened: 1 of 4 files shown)


orogvany:
screen shot 2018-05-17 at 8 21 41 pm

orogvany:
Good call! Thanks, will update

src/main/java/org/semux/gui/panel/TransactionsPanel.java
```diff
@@ -214,10 +315,29 @@ protected void refresh() {
...
-         * update table model
+         * update model
          */
         Transaction tx = getSelectedTransaction();
-        tableModel.setData(transactions);
+
+        // track all used addresses
+        for (StatusTransaction statusTransaction : transactions) {
+            Transaction transaction = statusTransaction.getTransaction();
+            to.add(transaction.getFrom());
```
cryptokat: typo here, `getFrom` should be `getTo`

orogvany: Oops! Thanks.

On the subject.  Do you think its worthwhile to filter the dropdowns based on other filters.  So if select vote the to/from will filter down to just those that have values?

cryptokat: I believe that would be quite useful for heavy users

orogvany: Ok, I'll take a stab at that.

src/main/java/org/semux/gui/panel/TransactionsPanel.java
```diff
@@ -182,7 +281,9 @@ public synchronized void actionPerformed(ActionEvent e) {
      * Refreshes this panel.
      */
     protected void refresh() {
-        List<StatusTransaction> transactions = new ArrayList<>();
+        transactions.clear();
+        Set<byte[]> to = new HashSet<>();
+        Set<byte[]> from = new HashSet<>();
```
semuxdev: `byte[]` doesn't work properly with `HashSet` because of the `hashcode()` method.

src/main/java/org/semux/gui/panel/TransactionsPanel.java
```diff
@@ -182,7 +281,9 @@ public synchronized void actionPerformed(ActionEvent e) {
      * Refreshes this panel.
      */
     protected void refresh() {
```
semuxdev: I was wondering if the `refresh()` would be heavy.

src/main/java/org/semux/gui/panel/TransactionsPanel.java
semuxdev: I'd like to request a `value` filter.

src/main/java/org/semux/gui/panel/TransactionsPanel.java
orogvany: Ah. I had this as ByteSet and reverted as it seemed to work without.  I think maybe it resolves itself when we map to ComboBoxItem.  I can move back to ByteSet

src/main/java/org/semux/gui/panel/TransactionsPanel.java
orogvany: It's heavier than it used to be, but I've noticed no real lag when testing with 4-5k transactions to filter.  Any ideas how to make it refresh more appropriately?

orogvany: I've tested it with 1000s of transactions across 100 accounts.  But in cases where one wallet contains 100s of addresses with 1000s of transactions, fixes for #830 will help keep this minimized.

containsAll() for a limited number of transactions (1k ~ 10k) is fairly trivial.  I would like to see more of a pubsub type refresh rather than on a timer.  I've made sure to only update where actual changes are seen, but there is room for improvement.

I think the utility of allowing filtering outweighs it, but yes, I think there's some ways to optimize for heavy accounts.

Sounds good.  I'll work on a min/max transaction amount filter.  That should not add any weight, as it wont be a dropdown that needs to be populated

orogvany: Let's hold off on merge.  I can add a WIP flag if we wish while I work through to minimize perf hit.  It doesn't feel excessive right now, but I think we can do better.

orogvany: I'm also wondering if pulling the filters out to its own class might clean this up a bit..  Seems like most panels just remain one class, but would be easier to manage as separate component..

orogvany:
screen shot 2018-05-19 at 4 29 42 pm

Ok, feeling a bit better about functionality.  Added value filtering too

src/main/java/org/semux/gui/panel/TransactionsPanelFilter.java
```diff
@@ -0,0 +1,209 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(TransactionsPanelFilter.class);
+
+    private transient final SemuxGui gui;
+
+    private final JComboBox<ComboBoxItem<TransactionType>> selectType;
+    private final JComboBox<ComboBoxItem<byte[]>> selectFrom;
+    private final JComboBox<ComboBoxItem<byte[]>> selectTo;
+    private final JTextField txtMin;
+    private final JTextField txtMax;
```
cryptokat: Would it be possible to use JSpinner instead and limit the input to numbers?

orogvany: With a JSpinner, you need to define a step size, so we'd have to figure out what that is.

E.g. if I type in 3.2 it'll floor it to 3.  I can set it to use a step of 0.1 but then it takes forever to move in any direction.

When trying to use the spinner, as users have different amounts, using it isn't really useful over just using a textbox.

PR #861, opened 2018-05-18
GUI: Add dropdown for selecting recipient

### LINKED ISSUE #479: Select a nicknamed recepeint from a dropdown on Send page
#### System information

Semux version: v1.0.0-rc.5

OS & Version: Windows 10 64 bit

#### Expected behaviour

There should be an option to select a receipent from the list in order to avoid mistakes in destination address.

#### Actual behaviour

No such functionality

It is an input field on Send page instead of a separate page

orogvany:
Add an editable dropdown for selecting account to send to.
Users can type in address as per old method, or select
either from their own wallet, or from anyone in address
book.

fixes #479


orogvany:
screen shot 2018-05-17 at 9 34 35 pm

orogvany:
I think i need one more fix.  Ive noted in many of our dropdowns (vote/etc) that when a refresh happens and you are interacting with popup things get dumb.  Either only update on change, pubsub, or block changes while being interacted with.  Not aure most common method for handling...

orogvany:
For the fix, went with checking values.  If we do want a more performant fix, we should just be able to update the list if the dropdown window isn't open.

If it's currently being interacted with, it may be preferrable to not update while they're in it ever.

But can remove WIP flag

orogvany:
Merged, thanks @semuxdev

PR #862, opened 2018-05-18
GUI: Support quoted parameters in console

orogvany:
For calls like signMessage and createAccount, we should
allow inputs with spaces.  This allows for users to enter
createAccount "Name with spaces"


orogvany:
We may want to go further an allow for escaped \"
Probably want to look at more robust tokenization in that case.

thoughts?

cryptokat:
Nashorn interpreter can be useful for tokenization by exposing the Semux API interface to Javascript: 
http://winterbe.com/posts/2014/04/05/java8-nashorn-tutorial/

PR #868, opened 2018-05-19
API: add 'GET /syncing' endpoint

cryptokat:
This API allows consumers to retrieve current sync status programmatically

### DIFF
--- a/src/main/java/org/semux/api/v2_0_0/impl/SemuxApiServiceImpl.java
+++ b/src/main/java/org/semux/api/v2_0_0/impl/SemuxApiServiceImpl.java
@@ -46,6 +46,7 @@ import org.semux.api.v2_0_0.model.GetLatestBlockResponse;
 import org.semux.api.v2_0_0.model.GetPeersResponse;
 import org.semux.api.v2_0_0.model.GetPendingTransactionsResponse;
 import org.semux.api.v2_0_0.model.GetRootResponse;
+import org.semux.api.v2_0_0.model.GetSyncingProgressResponse;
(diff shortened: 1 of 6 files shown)


semuxdev:
NOTE: syncing is triggered periodically after full sync. Application depending on the `syncing` property needs to be carefully designed.

src/main/java/org/semux/api/v2_0_0/impl/SemuxApiServiceImpl.java
```diff
@@ -744,6 +747,26 @@ public Response vote(String from, String to, String value, String fee) {
...
+    @Override
+    public Response getSyncingProgress() {
+        GetSyncingProgressResponse resp = new GetSyncingProgressResponse();
+        SyncingProgressType result = new SyncingProgressType();
+
+        if (kernel.getSyncManager().isRunning()) {
+            SyncManager.Progress progress = kernel.getSyncManager().getProgress();
+            result.setSyncing(true);
+            result.setStartingHeight(String.valueOf(progress.getStartingHeight()));
+            result.setCurrentHeight(String.valueOf(progress.getCurrentHeight()));
```
orogvany: We should probably set current height whether syncing is going on or not, else how are we expecting API users to use this?  Presumablyl they'd always want current height.

cryptokat: This endpoint enables a Parity-like web frontend to display current sync status. Current height is already available in the response of `GET /info` endpoint.

PR #873, opened 2018-05-20
Docs: Update docs boilerplate

orogvany:
Reorganize docs a little.  Add some consistent headers
and descriptions of what a page is to contain.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,4 +1,4 @@
-Welcome to the semux wiki!
+Welcome to the semux documentation!
 
 ### How it Works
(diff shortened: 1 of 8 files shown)


docs/README.md
```diff
@@ -1,11 +1,19 @@
 Welcome to the semux wiki!
```
cryptokat: Suggest to replace the `wiki` here with `documentation` as this is no longer a wiki.

orogvany: Ideally we'd have a real wiki that is open to easy edit, but githubs wikis are lacking a bit.

Yea, seems fair to update it.

PR #881, opened 2018-05-21
Provide Safe Ways to Automatically Unlock Wallet File

### LINKED ISSUE #878: semux.service template shouldn't have password included
in sample semux.service file, it shows 
ExecStart=/home/username/semux/semux-cli.sh --password "YOUR_WALLET_PASSWORD"

As users can see this with a ps command, it seems like we should allow for a better way to store this or apply so it does not show up in processes list.

cryptokat:
fix #878 

- [x] Add config property `wallet.password`
- [x] Add support of env variable `SEMUX_WALLET_PASSWORD`
- [x] Issue a warning message when the posix permission of semux.properties is not 600
- [x] Replace `--password` option with `SEMUX_WALLET_PASSWORD` in the systemd service unit template


src/main/java/org/semux/cli/SemuxCli.java
```diff
@@ -122,6 +125,10 @@ public void start(String[] args) throws ParseException, IOException {
 
         if (cmd.hasOption(SemuxOption.PASSWORD.toString())) {
             setPassword(cmd.getOptionValue(SemuxOption.PASSWORD.toString()));
+        } else if (getConfig().walletPassword() != null) {
+            setPassword(getConfig().walletPassword());
+        } else if (System.getenv(ENV_SEMUX_WALLET_PASSWORD) != null) {
+            setPassword(System.getenv(ENV_SEMUX_WALLET_PASSWORD));
```
semuxdev: Better document that configuration has higher priority high the environmental variable.

src/main/java/org/semux/cli/SemuxCli.java
```diff
@@ -163,8 +170,14 @@ protected void start() throws IOException {
...
-        if (SystemUtil.isPosix() && !wallet.isPosixPermissionSecured()) {
-            logger.warn(CliMessages.get("WarningWalletPosixPermission"));
+        if (SystemUtil.isPosix()) {
+            if (!wallet.isPosixPermissionSecured()) {
+                logger.warn(CliMessages.get("WarningWalletPosixPermission"));
+            }
+
+            if (!FileUtil.isPosixPermissionSecured(getConfig().getFile())) {
+                logger.warn(CliMessages.get("WarningConfigPosixPermission"));
+            }
```
semuxdev: Do we need to update the maven config to set the `config/semux.properties` file to proper perm for releasing?

cryptokat: It's done in 1.1 release: https://github.com/semuxproject/semux/commit/af8483b7a8b6d61db882adde3d5f83136d13c647

src/main/java/org/semux/cli/SemuxCli.java
cryptokat: Docs added in https://github.com/semuxproject/semux/pull/881/commits/d7d89ef316d5163b0cfb0e7b8d213d8c90b495ec

PR #883, opened 2018-05-21
API: Add Optional 'nonce' and 'validateNonce' Parameters to Transaction Ops of API 2.1

### LINKED ISSUE #880: Unable to set `nonce` for `/transfer`
I was trying to send transactions with a partially synced client. All my transactions got rejected, because I was unable to set the nonce for transactions, using the `/transfer` API.  The only way to achieve this is to use `/compose-raw-transaction`, which is not very convenient because it involves two API calls to get the transaction broadcast.

Suggest adding an optional `nonce` field to the existing `/transfer`, `/vote`, `/unvote`, `/delegate` endpoints.

cryptokat:
fix #880 

- [x] add `nonce` parameter, default to the current sender's nonce if omitted
- [x] add `validateNonce` parameter, default to true if omitted

### DIFF
--- a/src/main/java/org/semux/api/ApiHandlerImpl.java
+++ b/src/main/java/org/semux/api/ApiHandlerImpl.java
@@ -21,6 +21,7 @@ import java.util.stream.Collectors;
 import javax.ws.rs.QueryParam;
 
 import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
 import org.semux.util.exception.UnreachableException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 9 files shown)


src/main/java/org/semux/api/ApiHandlerImpl.java
```diff
@@ -115,22 +116,40 @@ private void loadRoutes() {
...
+                        }
+
+                        // convert params
+                        if (p.getRight().equals(Boolean.class)) {
+                            return Boolean.parseBoolean(param);
+                        }
+
+                        if (p.getRight().equals(Integer.class)) {
+                            return Integer.parseInt(param);
+                        }
```
semuxdev: Are we distinguishing `Integer` and `Long`?

src/main/java/org/semux/api/v2_1_0/impl/SemuxApiServiceImpl.java
```diff
@@ -607,7 +608,8 @@ public Response broadcastRawTransaction(String raw) {
 
             // tx nonce is validated in advance to avoid silently pushing the tx into
             // delayed queue of pending manager
-            if (tx.getNonce() != kernel.getPendingManager().getNonce(tx.getFrom())) {
+            if ((validateNonce == null || validateNonce)
+                    && tx.getNonce() != kernel.getPendingManager().getNonce(tx.getFrom())) {
```
semuxdev: Could we disable nonce validation by default? The pending manager may not reflect the actual nonce status of an account in the network, especially for application that manages nonce themselves.

src/main/java/org/semux/api/v2_1_0/impl/SemuxApiServiceImpl.java
```diff
@@ -814,18 +817,30 @@ public Response failure(Response.Status status, String message) {
...
+            }
+
+            Transaction tx = transactionBuilder.buildSigned();
+
+            // tx nonce is validated in advance to avoid silently pushing the tx into
+            // delayed queue of pending manager
+            if ((validateNonce == null || validateNonce)
+                    && tx.getNonce() != kernel.getPendingManager().getNonce(tx.getFrom())) {
+                return failure(resp, "Invalid transaction nonce.");
+            }
```
semuxdev: Same as before

src/main/java/org/semux/api/ApiHandlerImpl.java
cryptokat: removed as it's unused

src/main/java/org/semux/api/v2_1_0/impl/SemuxApiServiceImpl.java
cryptokat: Indeed disabling nonce check by default for customized nonces makes more sense.
`validateNonce` is disabled by default for tx endpoints in the latest commits.

FYI @witoldsz `POST /transaction/raw` no longer validates tx nonce when `validateNonce=true` is not provided.

PR #905, opened 2018-05-25
Sync speedup

kaar2:
Added faster sync. This new implementation basically does what was suggested here: in #803 with some modifications to handle invalid blocks. For each validator set, it first validates the votes for the last block, and then compares each block hash against it's child ParentHash field. If all block hashes are valid, they are added to the chain through validateApplyBlock method, while skipping the vote validation. If not all blocks in the validator sets have been forged, it will simply sync normally.

I did some benchmarks: in the original sync, the validateAndApply method was running in 28 miliseconds on average (~35 blocks per second), and in the new implementation it runs in 0.95 miliseconds on average (~1000 blocks per second). So the bottleneck is now in the download, and not in the validation (was able to sync ~340k blocks in ~35 mins, thats roughly 160 blocks per second, much less than the average speed of validateApplyBlock. 
Would be great if others could make benchmarks aswell, as I think it's possible to achieve even better results with a better connection than mine.

### DIFF
--- a/src/main/java/org/semux/consensus/SemuxSync.java
+++ b/src/main/java/org/semux/consensus/SemuxSync.java
@@ -258,7 +258,8 @@ public class SemuxSync implements SyncManager {
             Long task = toDownload.first();
 
             // quit if too many pending blocks
-            if (toProcess.size() > MAX_PENDING_BLOCKS && task > toProcess.first().getKey().getNumber()) {
+            int pendingBlocks = toProcess.size() + currentSet.size() + toFinalize.size();
+            if (pendingBlocks > MAX_PENDING_BLOCKS && task > toProcess.first().getKey().getNumber()) {
                 logger.trace("Pending block queue is full");
                 return;
             }
--- a/src/main/java/org/semux/consensus/SemuxSync.java
+++ b/src/main/java/org/semux/consensus/SemuxSync.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Random;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -110,6 +111,12 @@ public class SemuxSync implements SyncManager {
     private Map<Long, Long> toComplete = new HashMap<>();
     private TreeSet<Pair<Block, Channel>> toProcess = new TreeSet<>(
             Comparator.comparingLong(o -> o.getKey().getNumber()));
+    private TreeSet<Pair<Block, Channel>> currentSet = new TreeSet<>(
+            Comparator.comparingLong(o -> o.getKey().getNumber()));
+    private TreeMap<Long, Pair<Block, Channel>> toFinalize = new TreeMap<>();
+    
+    private long lastBlockInSet;
+    private boolean fastsync;
     private final Object lock = new Object();
 
     // current and target heights
@@ -140,6 +147,8 @@ public class SemuxSync implements SyncManager {
                 toDownload.clear();
                 toComplete.clear();
                 toProcess.clear();
+                currentSet.clear();
+                toFinalize.clear();
 
                 begin.set(chain.getLatestBlockNumber() + 1);
                 current.set(chain.getLatestBlockNumber() + 1);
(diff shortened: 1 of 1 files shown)


kaar2:
@cryptokat You are right of course. I forgot to mention that you should tune the value of MAX_UNFINISHED_JOBS. In my testing, I got best results when I set it between 200 and 250, but I believe the optimized value changes in different PC configurations. 
From my experience, values below a certain threshold cause constant halts in the sync every 200 blocks. Above that, the halts are pretty random, except that higher values seem to cause less frequent but longer halts (which kinda makes sense).

I guess tuning the value of MAX_PENDING_BLOCKS might provide better performance as well, but increasing it will also increase the memory usage, so I left it unchanged for now.

cryptokat:
@kaar2 Good work! The sync time decreased ~60% with little memory overhead after increasing `MAX_UNFINISHED_JOBS` and `MAX_PENDING_BLOCKS`.
selection_021

src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -314,64 +325,205 @@ private void process() {
...
+                            break;
+                        }
+                    }
+                    // Validate remaining block hashes
+                    iter = currentSet.descendingIterator(); 
+                    while (iter.hasNext()){
+                        Pair<Block, Channel> p = iter.next();
+                        if (p.getKey().getNumber() <= latest) {
+                            iter.remove();
+                        }
```
semuxdev: Isn't only blocks with `number > latest` added to the current set?

src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -314,64 +325,205 @@ private void process() {
...
+                        else if (!toFinalize.containsKey(p.getKey().getNumber())){
+                            Pair<Block, Channel> child = toFinalize.get(p.getKey().getNumber() + 1);
+                            if (child != null){
+                                if (Arrays.equals(child.getKey().getParentHash(), p.getKey().getHash())){
+                                    toFinalize.put(p.getKey().getNumber(), p);
+                                    iter.remove();
+                                }
+                                else{
+                                    handleInvalidBlock(p.getKey(), p.getValue());
```
semuxdev: How do we know if `p` is invalid or `child` is invalid?

src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -314,64 +325,205 @@ private void process() {
...
+     * @param channel 
+     */
+    protected void handleInvalidBlock(Block block, Channel channel){
+        InetSocketAddress a = channel.getRemoteAddress();
+        logger.info("Invalid block from {}:{}", a.getAddress().getHostAddress(), a.getPort());
+        toDownload.add(block.getNumber());
+        toComplete.remove(block.getNumber());
+        currentSet.remove(Pair.of(block, channel));
+        toFinalize.remove(block.getNumber(), Pair.of(block, channel));
+        toProcess.remove(Pair.of(block, channel));
```
semuxdev: Needs synchronization here

src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -314,64 +325,205 @@ private void process() {
...
      */
-    protected boolean validateApplyBlock(Block block) {
+    protected boolean validateApplyBlock(Block block, boolean validateVotes) {
         AccountState as = chain.getAccountState().track();
         DelegateState ds = chain.getDelegateState().track();
-
-        return validateBlock(block, as, ds) && applyBlock(block, as, ds);
+        boolean validated = validateVotes ? validateBlock(block, as, ds, true) : 
+                validateBlock(block, as, ds, false);
+        return validated && applyBlock(block, as, ds);
```
semuxdev: `return validateBlock(block, as, ds, validateVotes) && applyBlock(block, as, ds)`

src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -314,64 +325,205 @@ private void process() {
...
-                if (p.getKey().getNumber() <= latest) {
-                    iterator.remove();
-                } else if (p.getKey().getNumber() == latest + 1) {
-                    iterator.remove();
-                    pair = p;
-                    break;
-                } else {
-                    break;
+            // Validate votes only for the last block in each validator set.
+            // For each block in the set, compare it's hash against it's child hash.
```
orogvany: it's -> its

src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -314,64 +325,205 @@ private void process() {
...
+                        }
+                    }
+                    // Validate remaining block hashes
+                    iter = currentSet.descendingIterator(); 
+                    while (iter.hasNext()){
+                        Pair<Block, Channel> p = iter.next();
+                        if (p.getKey().getNumber() <= latest) {
+                            iter.remove();
+                        }
+                        else if (!toFinalize.containsKey(p.getKey().getNumber())){
```
orogvany: should run the 'mvn formatter:format license:format' to fix formatting violations

src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -314,64 +325,205 @@ private void process() {
...
+                                    toFinalize.put(p.getKey().getNumber(), p);
+                                    toComplete.remove(p.getKey().getNumber());
+                                }
+                                else{
+                                    handleInvalidBlock(p.getKey(), p.getValue());
+                                    break;
+                                }                                 
+                            }
+                            else{
+                                break;
```
orogvany: This is fairly deep sets of if blocks, some logging here at trace level would be appropriate/appreciated

orogvany: Additionally, some cleanup to extract methods into more bitesized sections would also be encouraged.

150+ line methods are usually not great for testing/maintenance. Some refactoring for maintainability would be appropriate IMO.

src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -314,64 +325,205 @@ private void process() {
...
+        InetSocketAddress a = channel.getRemoteAddress();
+        logger.info("Invalid block from {}:{}", a.getAddress().getHostAddress(), a.getPort());
+        toDownload.add(block.getNumber());
+        toComplete.remove(block.getNumber());
+        currentSet.remove(Pair.of(block, channel));
+        toFinalize.remove(block.getNumber(), Pair.of(block, channel));
+        toProcess.remove(Pair.of(block, channel));
+
+       // disconnect if the peer sends us invalid block
+       channel.getMessageQueue().disconnect(ReasonCode.BAD_PEER);
```
orogvany: This seems problematic/inconsistent.  In normal sync, if we get a bad message we do not disconnect from that peer.  We should be consistent.

src/main/java/org/semux/consensus/SemuxSync.java
orogvany: Unit tests would also be appreciated :D

kaar2:
I accidentally pushed the commits before testing them. Please ignore the new commits for now, I'll finish testing them and update the PR.

src/main/java/org/semux/consensus/SemuxSync.java
orogvany: Oh, sorry, I must have missed it.  Didn't see any other disconnects in this class

kaar2:
I've finished testing the new commits. Besides the small bug I fixed, everything seems to be working as expected. I would also like to see the effect of invalid blocks on the sync speed. Hopefully I'll find time to test it this weekend. Any more input will be much appreciated.

src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -315,16 +332,62 @@ private void process() {
...
+                }
 
+                // Validate remaining block hashes in current set
+                validateSetHashes();
+                return;
+            }
+
+            // Check if next block is ready to be applied to the chain
+            if (fastSync) {
+                iter = toFinalize.values().stream().sorted(Comparator.comparingLong(o -> o.getKey().getNumber()))
```
orogvany: Is there a reason this one is sorted while streaming, but the other 2 sets of values are sorted tree maps are sorted with comparator in constructor?  For consistency, might as well put comparator up in constructor as treemap is in order anyway, so this way ends up sorting twice.

src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -315,16 +332,62 @@ private void process() {
...
+                        currentSet.clear();
+                        lastBlockInSet = latest + config.getValidatorUpdateInterval();
+                        fastSync = true;
+                    }
+                }
+            }
+        }
+
         Pair<Block, Channel> pair = null;
+        Iterator<Pair<Block, Channel>> iter;
```
orogvany: Might as well keep the name 'iterator' rather than change to 'tier'.  Clarity over brevity.

src/main/java/org/semux/consensus/SemuxSync.java
orogvany: It seems a little unnecessary to reuse the 'iter' variable here, since they seem to be used in different fashions.  It's used for toProcess at the top, then the iterator is reinitialized with possibly different.  Naming is important  /pedantic off

src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -333,45 +396,114 @@ private void process() {
...
 
-        if (pair != null) {
-            logger.info("{}", pair.getKey());
-
-            if (validateApplyBlock(pair.getKey())) {
-                synchronized (lock) {
+        // Validate and apply block to the chain
+        synchronized (lock) {
+            if (pair != null) {
+                logger.info("{}", pair.getKey());
```
orogvany: Might want to give context to what is being logged here, rather than just number alone

src/main/java/org/semux/consensus/SemuxSync.java
```diff
@@ -333,45 +396,114 @@ private void process() {
             }
         }
 
-        if (pair != null) {
-            logger.info("{}", pair.getKey());
-
-            if (validateApplyBlock(pair.getKey())) {
-                synchronized (lock) {
+        // Validate and apply block to the chain
+        synchronized (lock) {
```
orogvany: We may want to make sure the synchronization here is most optimal.  It looks like we've moved from just set add/removes in sync blocks to larger sets of logic and validation inside a single threaded lock.  No a problem per-say, but worth double checking

kaar2:
@orogvany Thanks for the input. The sorting of `toFinalize` was indeed redundant. I removed it and gave clearer names to the iterators. About the logging, `pair.getKey()` returns the actual block, not the block number. So what will actually be logged is the `toString()` method of the block.
Finally, I'm aware of the larger synchronization blocks in this implementation, but I don't see much more room to play here. I did reduce the size of one block, so it won't include the call to `handleInvalidBlock`, but there are other calls to that method inside `validateSetHashes`. If it's a big concern, I can move the peer disconnection outside that method.

PR #906, opened 2018-05-25
GUI: fix account ordering on select from account

orogvany:
Select account uses array ordering.
This fixes the 1.2.0 branch and should be considered and emergency fixpack

The RC will select the wrong account.

### DIFF
--- a/src/main/java/org/semux/gui/panel/SendPanel.java
+++ b/src/main/java/org/semux/gui/panel/SendPanel.java
@@ -432,9 +432,11 @@ public class SendPanel extends JPanel implements ActionListener {
      */
     protected WalletAccount getSelectedAccount() {
         AccountItem selected = (AccountItem) selectFrom.getSelectedItem();
-        for (WalletAccount account : model.getAccounts()) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/semux/gui/panel/SendPanel.java
```diff
@@ -431,8 +431,16 @@ protected void showAddressBook() {
...
+        AccountItem selected = (AccountItem) selectFrom.getSelectedItem();
+        if (selected != null) {
+            for (WalletAccount account : model.getAccounts()) {
+                if (Arrays.equals(selected.address, account.getAddress())) {
+                    return account;
+                }
+            }
+        }
+
+        return null;
```
cryptokat: suggest to use stream api
```java
    protected WalletAccount getSelectedAccount() {
        AccountItem selected = (AccountItem) selectFrom.getSelectedItem();
        return selected == null ? null : model.getAccounts()
                .stream()
                .filter(walletAccount -> Arrays.equals(selected.address, walletAccount.getAddress()))
                .findFirst()
                .orElse(null);
    }
```

PR #915, opened 2018-06-04
Net: Ntp check

orogvany:
When we do version checks, also check that the NTP time
is in sync and warn the user that their system time
needs updating.


witoldsz:
I was thinking, that instead of warning the peer to fix it's time, just fetch the current time from some well known NTP server, compare it with local time and apply the difference when working with other peers, so they all are operating without that annoying problem?

orogvany:
Yea, I can do that.  We will need to be mindful in future of System.currentTimeMillis creeping back in though, so it adds some risk

